PR #2, opened 2017-08-06
Add tests

rsoeldner:
@jmcardon please review , the `should generate and verify with default settings` test fails.


jmcardon:
Passwordhasher tests should only really needed to test if you can encrypt/decrypt.

If you want to reverse engineer the steps, you need to separate the parameters encoded in the password and _then_ verify. I.e BCrypt stores log rounds, salt and password. You can extract each one to check the password, as changing any of these changes the computation of the password.

Thank you for your hard work :+1: .

PR #3, opened 2017-08-08
[WIP] Add AES and DES symmetric Keygenerator

rsoeldner:


src/main/scala/fucc/cipher/a.sc
```diff
@@ -0,0 +1,20 @@
...
+val plain: ClearText = ClearText("abcdef".getBytes)
+
+
+val x = for {
+  enc <- `DES/ECB/PKCS5Padding`.encrypt(plain)
+  dec <- `DES/ECB/PKCS5Padding`.decrypt(enc)
+} yield dec
+
+x match {
+  case Right(msg) => println(s"Wroks : ${msg.content}")
```
rsoeldner: At this point I wasn't able to recover the plain object (after encryption & decryption)

src/main/scala/fucc/cipher/instance.scala
```diff
@@ -0,0 +1,32 @@
...
+import java.security.SecureRandom
+import javax.crypto.spec.IvParameterSpec
+import javax.crypto.{SecretKey, Cipher => JCipher}
+
+object instance {
+
+  val `DES/ECB/PKCS5Padding` = new Cipher[DES, ECB, `PKCS5Padding`] {
+
+    def generator = JCipher.getInstance(s"${DES.tag.algorithm}/${ECB.tag.algorithm}/${`PKCS5Padding`.tag.algorithm}")
+    def getIV(): Array[Byte] = List(0,0,0,0,0,0,0,0).map(_.toByte).toArray
```
rsoeldner: Several Stackoverflow entries mentioned to specify the IvParameterSpec otherwise I would receive an Error.

src/main/scala/fucc/cipher/instance.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    override def encrypt(clearText: ClearText)(implicit key: SecretKey): Either[CipherError, CipherText] = {
```
rsoeldner: Maybe we should wrap the key object for each algorithm otherwise you can compile passing an AES Key to a DES Cipher.

jmcardon: You have to be careful here. IV is an initialization vector, and depends on your cipher instance.

getIV depends on the cipher itself. You would call it on an instance of Cipher, 

In general, make getIV() take the cipher instance.

src/main/scala/fucc/cipher/a.sc
jmcardon: I will take a look

jmcardon:
https://gist.github.com/praseodym/f2499b3e14d872fe5b4a

Funny enough, we need to abstract out the AlgorithmParameterSpec as well, unfortunately.

Not all implementations take IvParameterSpec

src/main/scala/tsec/cipher/instances/JCASymmetricCipher.scala
```diff
@@ -97,11 +131,32 @@ class JCASymmetricCipher[A, M, P](
...
+    * @tparam P Padding mode
+    * @return
+    */
   def getCipher[A: SymmetricAlgorithm, M: ModeKeySpec, P: Padding]
     : Either[NoSuchInstanceError, JCASymmetricCipher[A, M, P]] = {
     val c = new JCASymmetricCipher[A, M, P]
     c.genInstance.map(_ => c).leftMap(_ => NoSuchInstanceError)
   }
+
+  /**
```
rsoeldner: <3

PR #11, opened 2017-08-16
(WIP) Symmetric cleanups and hmac

jmcardon:
loads of goodies

Initial implementation for Mac Algebras.

There needs to be some cleanup around CipherError: Some of those errors are more general than was expected initially, so either we change some stuff to be `SymmetricKeyError` or something, or we propagate ciphererror.

Ideas would be welcome on how to do this

The macalgebra needs to get rid of throwable, but of course, we need an improved error type to even do this.


src/main/scala/tsec/mac/instance/HMACSHA384.scala
```diff
@@ -0,0 +1,3 @@
+package tsec.mac.instance
+
+class HMACSHA384
```
rsoeldner: What happened here ?
Missing implementation ?

jmcardon: yes

PR #18, opened 2017-10-03
(WIP) 0.0.1-M2 

jmcardon:
Breaking changes:

- The syntax classes from `tsec.common.ByteUtils._` were moved to the package object of tsec.common. Thus, the import is now `import tsec.common._`
- `DefaultEncryptor` renamed to just `Encryptor`, but this did not change the name of the object `DefaultEncryptor`
- Same for auth encryptor

Internal Non-breaking changes:
- Moved every wrapped array type into an existential newtype (Thanks edmund!).

New Features :tada: :
- [x] A bit more syntactic sugar!
- [x] Http4s signed cookies
- [x] Http4s AEAD cookies
- [ ] Http4s JWT Stuff


jmcardon:
You're right... It's really messy. This is my mistake. The types overhaul should've been a PR of it's own.

I'll start merging into this branch.

PR #24, opened 2017-10-04
WIP Add assymetric cipher

rsoeldner:
with RSA 2048, 3072 and 4096

### DIFF
--- a/cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/core/AsymmetricCipherAlgebra.scala
+++ b/cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/core/AsymmetricCipherAlgebra.scala
@@ -1,0 +1,15 @@
+package tsec.cipher.asymmetric.core
+
+import tsec.cipher.asymmetric.imports._
+import tsec.cipher.common.mode.NoMode
+import tsec.cipher.common.{CipherText, PlainText}
+
+trait AsymmetricCipherAlgebra[F[_], A, P] {
+  type C
+
+  def genInstance: F[C]
+
+  def encrypt(plainText: PlainText, key: PublicKey[A]): F[CipherText[A, NoMode, P]]
+
+  def decrypt(cipherText: CipherText[A, NoMode, P], key: PrivateKey[A]): F[PlainText]
+}
--- a/cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/imports/JCAAsymmetricCipher.scala
+++ b/cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/imports/JCAAsymmetricCipher.scala
@@ -24,18 +24,18 @@ class JCAAsymmetricCipher[A, P](implicit algoTag: AsymmetricAlgorithm[A], paddin
   protected[this] def initEncryptor(cipher: JCipher, publicKey: PublicKey[A]): Either[CipherKeyError, Unit] =
     Either
       .catchNonFatal(
-        cipher.init(JCipher.ENCRYPT_MODE, publicKey)
+        cipher.init(JCipher.ENCRYPT_MODE, PublicKey.toJavaPublicKey[A](publicKey))
       )
       .mapError(CipherKeyError.apply)
 
   protected[this] def initDecryptor(
       cipher: JCipher,
-      publicKey: PrivateKey[A],
+      privateKey: PrivateKey[A],
       iv: Array[Byte]
   ): Either[CipherKeyError, Unit] =
     Either
       .catchNonFatal(
-        cipher.init(JCipher.DECRYPT_MODE, publicKey)
+        cipher.init(JCipher.DECRYPT_MODE, PrivateKey.toJavaPublicKey[A](privateKey))
       )
       .mapError(CipherKeyError.apply)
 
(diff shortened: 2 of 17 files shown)


cipher-symmetric/src/main/scala/tsec/cipher/symmetric/imports/package.scala
```diff
@@ -12,13 +12,13 @@ package object imports{
...
 
   /**
-   * Typeclass for propagating symmetric key algorithm information
-   * Note: Key length is in bits
-   *
-   * @param algorithm the symmetric cipher representation, as a string
-   * @param keyLength key length in bits
-   * @tparam T Parametrized cipher type
-   */
+    * Typeclass for propagating symmetric key algorithm information
```
ChristopherDavenport: Java or Scala Comment Blocks? I've seen this switch more then once now I think.

rsoeldner: @ChristopherDavenport I didn't know the difference, is it _only_ the first line, scala starts with /** Typeclass .... ?

cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/core/AsymmetricCipherAlgebra.scala
```diff
@@ -0,0 +1,15 @@
+package tsec.cipher.asymmetric.core
+
+import tsec.cipher.asymmetric.imports._
+import tsec.cipher.common.mode.NoMode
+import tsec.cipher.common.{CipherText, PlainText}
+
+trait AsymmetricCipherAlgebra[F[_], A, P] {
+  type C
```
ChristopherDavenport: Why internal typing here over a typed announced into the trait? As it appears only to be used in a location that would easily lend to this.

`trait AsymmetricCipherAlgebra[F[_], A, P, C]`

`extends AsymmetricCipherAlgebra[Either[CipherError, ?], A, P, JCipher]` 

However I'm still exploring the code base, so perhaps I just dont' understand yet.

jmcardon: @ChristopherDavenport this is more my design decision, as I wanted to maintain the cipher's representation as an internal thing. This algebra is very similar to [this](https://github.com/jmcardon/tsec/blob/c3566c48027035b2b7b0a1346e77e64a721a3e85/cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/SymmetricCipherAlgebra.scala).

In particular, it's mostly because I wanted to not expose the actual representation of the internal state of the cipher in the type signature, but treat it as sort of an "internal state type" that wouldn't usually escape to the outside world, unless you absolutely want it for whatever reason.

jmcardon: The logic behind it, mostly, is that if we ever decide to be polymorphic about the cipherAlgebra we're using, i.e we want to be polymorphic over `F[_], A, M, P`, I don't want a consumer (or myself) to have to pass the internal representation type to something that consumes it.

cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/imports/imports.scala
```diff
@@ -0,0 +1,48 @@
...
+  protected[tsec] case class AsymmetricAlgorithm[T](algorithm: String, keyLength: Int) extends CryptoTag[T]
+
+  sealed case class KeyPair[A](privateKey: PrivateKey[A], publicKey: PublicKey[A])
+  object KeyPair {
+    def apply[A](jPrivateKey: JPrivateKey, jPublicKey: JPublicKey): KeyPair[A] =
+      new KeyPair[A](PrivateKey[A](jPrivateKey), PublicKey[A](jPublicKey))
+  }
+
+  sealed trait TaggedKey[T] {
+    type KeyRepr[A] <: Id[T]
```
jmcardon: Just following up on our convo:
Remember, you cannot carry proof that KeyRepr[A] === T

Otherwise, the compiler reduces it down immediately, and the type segregation we want is lost.

Please separate taggedkey into two things: PrivateKey and PublicKey (alternatively copy it from signature).

cipher-asymmetric/src/main/scala/tsec/cipher/asymmetric/imports/JCAAsymmetricCipher.scala
```diff
@@ -0,0 +1,68 @@
...
+import tsec.common.ErrorConstruct._
+
+class JCAAsymmetricCipher[A, P](implicit algoTag: AsymmetricAlgorithm[A], paddingTag: Padding[P])
+    extends AsymmetricCipherAlgebra[Either[CipherError, ?], A, P] {
+
+  type C = JCipher
+
+  def genInstance =
+    Either
+      .catchNonFatal(JCipher.getInstance(s"${algoTag.algorithm}/None/${paddingTag.algorithm}", "BC"))
```
jmcardon: This worries me.

ECIES is a little different than AES and whatnot. i.e [this](https://github.com/bcgit/bc-java/blob/master/prov/src/test/java/org/bouncycastle/jce/provider/test/ECIESTest.java#L139) or: [this](https://github.com/bcgit/bc-java/blob/master/prov/src/test/java/org/bouncycastle/jce/provider/test/ECIESTest.java#L184)

Do you think you could verify that Cipher.getInstance("ECIES", "BC") returns the same as what is in your interpreter? [This](https://github.com/google/wycheproof/blob/master/java/com/google/security/wycheproof/testcases/EciesTest.java#L122) states what you may be inputting could be an invalid name.

jmcardon:
closing since it's been open forever, LOL.

On a serious note: Until someone absolutely needs to do RSA directly using tsec or something (which is strongly not recommended), I'd rather they use libsodium's edwards curve asymmetric crypto for it.

PR #26, opened 2017-10-05
Change comment style to scala-style

rsoeldner:
.. work for monkys :tada:


jmcardon:
Thanks so much @rsoeldner !

PR #43, opened 2017-11-10
Compose authenticators

jmcardon:
Relatively big PR, I apologize. It breaks a few things:
* `Authenticator` renamed to `AuthenticatorService`
* Authenticator is now a common trait that all tsec authenticators implement.
* `JWTAuthenticator` now needs a `AugmentedJWT` backing store instead of one for `JWTMAC`
* `JWTAuthenticator` now defaults to transporting as a bearer token 
* (Invisible to consumer) `JWTInternal` is gone.  No need for funky encoding/decoding anymore
* `CookieAuthenticator` (Signed cookies) changed name, it is now `SignedCookieAuthenticator`.
* Cookie-based authenticators dropped reliance on `HttpDate`, instead using `java.time.Instant`
* `AuthenticatorService` is now an abstract class...

Cool things this PR adds:
* `AuthenticatorService` are now Composable! You lose refined type information, but this solves the problem of multiple endpoints not being able to authenticate with one common interface, while still retaining _some_ common information. This is tested
* ~~`tryExtractRaw` added~~ `extractRawOption` added to AuthenticatorService cc @ChristopherDavenport 
* `parseRaw` added to AuthenticatorService
* `extractAndValidate` is implemented in terms of `extractRawOption` and `parseRaw` for all Authenticators.
This means you can make decisions on whether or not to proceed with authentication based on whether the necessary conditions exist for it to be present in raw format.
* `liftService` added to `SecuredRequestHandler`, so people can choose to compose `TSecAuthService`.
* JWTMAC fixed @aeons.
* JWTAuthenticator can now transport as a bearer token, or as in an arbitrary header.
* Base64Url encoding/decoding, in general, to use the padding-less version.
* TSec CSRF api has equal naming with http4s.

*edit*
I apologize for the large PR, but I wasn't comfortable merging this while I worked actively on the composability feature, and somewhere along the way, I decided to stuff in the `JWTAuthenticator` changes too. This should've been two PRs, one for the JWTAuthenticator changes, another for the composability feature.

### DIFF
--- a/tsec-http4s/src/main/scala/tsec/authentication/ECookieAuthenticator.scala
+++ b/tsec-http4s/src/main/scala/tsec/authentication/ECookieAuthenticator.scala
@@ -5,6 +5,7 @@ import java.util.UUID
 
 import cats.MonadError
 import cats.data.OptionT
+import cats.effect.Sync
 import io.circe.{Decoder, Encoder}
 import io.circe.parser.decode
 import org.http4s._
@@ -18,7 +19,8 @@ import tsec.cookies._
 import scala.concurrent.duration.FiniteDuration
 import io.circe.syntax._
 import io.circe.generic.auto._
-import cats.implicits._
+import cats.syntax.all._
+import cats.instances.string._
 import tsec.jwt.JWTPrinter
 
 sealed abstract class ECookieAuthenticator[F[_], I, V, A](implicit auth: AuthEncryptor[A])
@@ -155,7 +157,7 @@ object ECookieAuthenticator {
       tokenStore: BackingStore[F, UUID, AuthEncryptedCookie[A, I]],
       identityStore: BackingStore[F, I, V],
       key: SecretKey[A]
-  )(implicit M: MonadError[F, Throwable]): StatefulECAuthenticator[F, I, V, A] =
+  )(implicit F: Sync[F]): StatefulECAuthenticator[F, I, V, A] =
     new StatefulECAuthenticator[F, I, V, A](settings.expiryDuration, settings.maxIdle) {
 
       /** Return a new instance with a modified key */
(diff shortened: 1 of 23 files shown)


ChristopherDavenport:
Is this complete as is, or is there more coming?

tsec-http4s/src/main/scala/tsec/authentication/SCookieAuthenticator.scala
```diff
@@ -158,6 +150,9 @@ object CookieAuthenticator {
       ): OptionT[F, Unit] =
         if (validateCookie(internal, raw, now)) OptionT.pure[F](()) else OptionT.none
 
+      def tryExtractRaw(request: Request[F]): Option[String] =
```
ChristopherDavenport: nit: Naming convention - try method returning option `extractRawOption`?

tsec-http4s/src/main/scala/tsec/authentication/ECookieAuthenticator.scala
```diff
@@ -438,7 +444,7 @@ object EncryptedCookieAuthenticator {
         * @return
         */
       def discard(authenticator: AuthEncryptedCookie[A, I]): OptionT[F, AuthEncryptedCookie[A, I]] =
-        OptionT.pure(authenticator.copy[A, I](content = AEADCookie[A]("invalid"), expiry = HttpDate.now))
+        OptionT.pure(authenticator.copy[A, I](content = AEADCookie[A]("invalid"), expiry = Instant.now()))
```
ChristopherDavenport: `Instant.now` as a side-effect in pure method.

tsec-http4s/src/main/scala/tsec/authentication/SCookieAuthenticator.scala
jmcardon: Hmm... makes sense

tsec-http4s/src/main/scala/tsec/authentication/ECookieAuthenticator.scala
jmcardon: Ah, this is true.. I should change this everywhere though. Great catch

jmcardon:
Alright, I'm pretty satisfied with this. I'm sorry this pr was messy @ChristopherDavenport  @rsoeldner , but I went back and forth a lot on what I believed was a better interface for users.

Ultimately, my goal was to try to open up what you can do with authenticators and providing safer defaults. While at the same time, allowing more flexibility, such as branching decisions similar to what is implemented in [tailRecAuth](https://github.com/jmcardon/tsec/blob/3353a73649bd50347d203f83f5f86b8ffa5581ae/tsec-http4s/src/main/scala/tsec/authentication/AuthenticatorService.scala#L118) in regards to composability. 

As far as JWT, the `Authorization` header is a much safer transport default for `JWTAuthenticator` than an arbitrary header, as http4s on purpose redacts such a header from logging. On top of this, I decided to utilize the `subject` field more liberally, to what is ultimately the same outcome, but with less classes in between storing things statelessly in a secure manner.

Aside from this, plus a quick JWT fix, this is all this PR has in mind.

rsoeldner:
:+1:

Issue #58, opened 2017-11-23
Enhance password hashers to take char[] instead of string


jmcardon:

jmcardon:
done in #60

PR #60, opened 2017-12-10
(WIP) 0.0.1-M6 Changes.

jmcardon:
This is a rather large PR which breaks a load of things (for the greater good).

* Password hashing now requires an effect, which is proper considering that `SecureRandom` is used.
* Password Hashing has received a slight speed optimization for BCrypt via our `ManagedRandom`, and `StringBuilder` over `StringBuffer`, especially considering no synchronization is needed.
* #58 Byte arrays are now the default, and they are wiped after password hashing

Type signatures are now consistent over All modules. This means types have gone from:
* For message authentication codes: `A` to `MAC[A]` (i.e `HMACSHA256` to `MAC[SHA256]`)
* For digest hashes: `A` to  `CryptoHash[A]`  (i.e `SHA256` to `CryptoHash[SHA256]`)
* For password hashes: `A` to `PasswordHash[A]`
* For signatures: `A` to `CryptoSignature[A]`

Due to these type sigs being consistent, a few code smells have gone away: In particular, `StringEV` and `ByteEV` are no longer necessary, and thus have been removed. This has no impact on the user level api.

JWT received a few changes as well:
* JWT methods now default to monadic. This makes sense, as `tsec` is an FP-first library. Thus now, all `JWTMacM` methods have become `JWTMac` by default, and `JWTMac` has become `JWTMacImpure`, same with `JWTSig`.


common/src/main/scala/tsec/common/ManagedRandom.scala
```diff
@@ -13,7 +13,7 @@ trait ManagedRandom {
     * https://tersesystems.com/2015/12/17/the-right-way-to-use-securerandom/
     * Random may block on linux machiens
     */
-  private var cachedRand: SecureRandom = {
+  private[tsec] var cachedRand: SecureRandom = {
```
jmcardon: this allows us to pass our cached rand in places where it takes the actual rand internally, then use `forceIncrement`

common/src/main/scala/tsec/common/ManagedRandom.scala
```diff
@@ -32,6 +32,12 @@ trait ManagedRandom {
     cachedRand = tmpRand
   }
 
+  private[tsec] def forceIncrement: Unit = {
```
jmcardon: This mutable behavior is private because most of the time, people should be using `nextBytes`.

calling forcedIncrement is only for when we pass the mutable reference around, with care.

PR #62, opened 2017-12-11
Improve api

jmcardon:
`tsec` is FP-first, thus JCA primitives should be _pure by default_, thus:

`JCAMacPure` => `JCAMac`
`JCASignerPure` => `JCASigner`

Also, non-api breaking: we now have helper methods for `Mac` and `Signature`!


signatures/src/main/scala/tsec/signature/imports/JCASigner.scala
```diff
@@ -1,24 +1,46 @@
...
-    extends SignaturePrograms[SigErrorM, A] {
+sealed abstract case class JCASigner[F[_]: Sync, A: SigAlgoTag](
+    alg: JCASigInterpreterPure[F, A]
+) extends SignaturePrograms[F, A] {
 
   type PubK  = SigPublicKey[A]
   type PrivK = SigPrivateKey[A]
   type Cert  = SigCertificate[A]
-  val algebra: JCASigInterpreterImpure[A] = alg
+  val algebra: JCASigInterpreterPure[F, A] = alg
```
hrhino: Why is this still `...Pure`?

(Also git's move detection is truly trash here...)

jmcardon: Good point. There's still a few `Pure` as far as the interpreters, but those shouldn't be referenced by the user. 

That said, good catch, I'll make it consistent.

PR #69, opened 2017-12-13
add note on binbreaking

jmcardon:


jmcardon:
Good point. I can add a note on that. Major versions = no bincompat. Minor versions = bincompat.

`tsec-0.1.0` shouldn't be bincompat with `tsec-0.2.0`, but `tsec-0.1.1` should be with `0.1.2`

ChristopherDavenport:
Isn't `0.0.1` patch, `0.1.0` minor, and `1.0.0` major? I feel I read that `0.1.0` was a major version which was odd to me.

README.md
```diff
@@ -20,6 +20,13 @@ For the current progress, please refer to the [RoadMap](https://github.com/jmcar
 For version changes and additions, including breaking changes, see either [release notes](https://github.com/jmcardon/tsec/releases)
 or the [Version Changes](https://github.com/jmcardon/tsec/wiki/Version-Changes) page.
 
+### Note on milestones:
+All `x.x.x-Mx` releases are milestone releases. Thus, we do not guarantee binary compatibility or no api-breaking until
+a concrete version(i.e `0.0.1`). We aim to keep userland-apis relatively stable, but 
+internals shift as we find better/more performant abstractions.
+
+We will guarantee compat between minor versions (i.e 0.0.1 => 0.0.2) but not major versions (0.0.1 => 0.1.0)
```
ChristopherDavenport: `s/minor/patch/g` `s/major/minor/g`

hrhino:
That's semver, yes. I'm not a huge fan of it as it is. I prefer to keep the leftmost digit for redesigns / huge changes or whatever: things that would require more than minor source changes. Having to bump that number just to break bincompat in any way seems unpleasant to me.

README.md
hrhino: good catch. reviewing on the train considered not that effective.

ChristopherDavenport:
All good, just thought if we were not going to keep semver we should note that explicitly, not merely describing how we are going to do it differently.

hrhino:
That's reasonable, although I think the strict definition of "semantic versioning" isn't too common in java/scala circles.

jmcardon:
@ChristopherDavenport you have a good point actually...

We should note on this. I have the leftmost digit in my mind as the "deploy the nukes, we have a huge change now" button, rather then semver.

hrhino:
That's because we're scala devs, and scala `3.x.y` involves a complete rewrite.

Also scalaz devs, for whom scalaz `8.x.y` involves a complete rewrite :P

README.md
```diff
@@ -20,6 +20,21 @@ For the current progress, please refer to the [RoadMap](https://github.com/jmcar
...
+Our Notation for versions is:
+```
+X.X.X
+^ ^ ^____Minor
+| |______Major
+|________Complete redesign (i.e scalaz 7 vs 8)  
+```
+
+All `x.x.x-Mx` releases are milestone releases. Thus, we do not guarantee binary compatibility or no api-breaking until
+a concrete version(i.e `0.0.1`). We aim to keep userland-apis relatively stable, but
```
hrhino: wee nit, missing space and period, i.e. "version (i.e. `0.0.1`)".

README.md
```diff
@@ -20,6 +20,21 @@ For the current progress, please refer to the [RoadMap](https://github.com/jmcar
...
+### Note on milestones:
+Our Notation for versions is:
+```
+X.X.X
+^ ^ ^____Minor
+| |______Major
+|________Complete redesign (i.e scalaz 7 vs 8)  
+```
+
+All `x.x.x-Mx` releases are milestone releases. Thus, we do not guarantee binary compatibility or no api-breaking until
```
hrhino: "no api-breakage" while I'm being a pedant

README.md
```diff
@@ -20,6 +20,21 @@ For the current progress, please refer to the [RoadMap](https://github.com/jmcar
...
+a concrete version(i.e `0.0.1`). We aim to keep userland-apis relatively stable, but 
+internals shift as we find better/more performant abstractions.
+
+We will guarantee compat between minor versions (i.e 0.0.1 => 0.0.2) but not major versions (0.0.1 => 0.1.0)
```
hrhino: Maybe just say "compatibility"?

PR #70, opened 2017-12-14
Replace RuntimeException

rsoeldner:


tsec-libsodium/src/main/scala/tsec/libsodium/ScalaSodium.scala
```diff
@@ -53,7 +53,7 @@ object ScalaSodium
     val sodium = new ScalaSodium()
     checkVersion(SodiumJNI.sodium_version_string.asInstanceOf[String])
     if (sodium.sodium_init < 0)
-      throw new RuntimeException("ScalaSodium is not safe to use")
+      throw new SodiumLoadError("ScalaSodium is not safe to use, sodium_init < 0")
```
hrhino: The [docs](https://download.libsodium.org/doc/usage/) claim that

> `sodium_init()` returns `0` on success, `-1` on failure, and `1` if the library had already been initialized

so the extra info about `sodium_init < 0` is a bit extraneous.

That said, I'm not so hot on the phrasing "`ScalaSodium` is not safe to use". Perhaps "libsodium failed to initialize"?

jmcardon: This is true... Maybe "Has not been initialized properly" is the proper way.

Maybe if `sodium_init == 1` we can give some info, though we still lack logging in gen

hrhino: Shrug (re: returning `1`). I think adding it to the logging wishlist might be good. 

On the other hand, it's not something that people should be worried about at all, so it could wind up just being noise.

PR #73, opened 2017-12-15
Add initial key derivation

rsoeldner:


rsoeldner:
I can't figure out why travis is failing, for me this locally passes all tests.
Another point is passing the sodium generated files, shouldn't it be enough to add the `sodium.i` file ?

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
```diff
@@ -0,0 +1,37 @@
...
+    val masterKey = PrivateKey[KeyDerivation](new Array[Byte](ScalaSodium.crypto_kdf_KEYBYTES))
+    S.crypto_kdf_keygen(masterKey)
+    masterKey
+  }
+
+  def deriveKey[F[_]](
+      masterKey: PrivateKey[KeyDerivation],
+      keyLength: Int,
+      context: Array[Byte]
+  )(implicit F: Sync[F], S: ScalaSodium): StateT[F, Int, SodiumKey[KeyDerivation]] =
```
jmcardon: You need to have the ability to get the same id twice. I don't think this helper is necessary.

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
```diff
@@ -0,0 +1,37 @@
...
+  def generateKey[F[_]](implicit F: Sync[F], S: ScalaSodium): F[PrivateKey[KeyDerivation]] = F.delay {
+    val masterKey = PrivateKey[KeyDerivation](new Array[Byte](ScalaSodium.crypto_kdf_KEYBYTES))
+    S.crypto_kdf_keygen(masterKey)
+    masterKey
+  }
+
+  def deriveKey[F[_]](
+      masterKey: PrivateKey[KeyDerivation],
+      keyLength: Int,
+      context: Array[Byte]
```
jmcardon: already said it over voice, but this should be `String`

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
```diff
@@ -0,0 +1,37 @@
...
+      keyLength: Int,
+      id: Int,
+      context: String
+  )(implicit F: Sync[F], S: ScalaSodium): F[SodiumKey[KeyDerivation]] = F.delay {
+    if (ScalaSodium.crypto_kdf_BYTES_MIN > keyLength || keyLength > ScalaSodium.crypto_kdf_BYTES_MAX)
+      throw KeyLengthError(keyLength)
+
+    val ctx = context.utf8Bytes
+    if (ctx.length != ScalaSodium.crypto_kdf_CONTEXTBYTES)
+      throw ContextBytesError(ctx.length)
```
hrhino: I know it's caught by the `delay` call, but I still dislike using `throw` in general. `Sync` (`ApplicativeError`) has `raiseError` for things like this, IIRC.

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/package.scala
```diff
@@ -0,0 +1,19 @@
+package tsec.libsodium
+
+package object kdf {
+
+  case class KeyLengthError(n: Int) extends Exception with Product with Serializable {
```
hrhino: surely these are `TSecErrors` now? :D

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/package.scala
```diff
@@ -0,0 +1,19 @@
+package tsec.libsodium
+
+package object kdf {
```
hrhino: why do the errors classes live in the package object? they'd be perfectly-fine top level decls.

tsec-libsodium/src/test/scala/tsec/libsodium/KeyDerivationTest.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  it should "generate multiple keys" in {
+
+    val program = for {
+      master <- KeyDerivation.generateKey[IO]
+      context = "Examples"
+      _ <- KeyDerivation.deriveKey[IO](master, 16, 1, context)
+      _ <- KeyDerivation.deriveKey[IO](master, 32, 2, context)
+      _ <- KeyDerivation.deriveKey[IO](master, 64, 3, context)
+    } yield ()
```
hrhino: What is this really testing? It doesn't assert anything about the derived keys other than that nothing blows up unexpectedly.

I'd expect (if I understand the API correctly) that

* `deriveKey` gives the same result for multiple invocations with the same arguments
* `deriveKey` gives different results (almost certainly) for a different context

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
rsoeldner: You are right, but the signature is `raiseError[A](e: Throwable): F[A]` and we wrapped it inside `F.delay`, I think there is no way arround ?

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/package.scala
rsoeldner: right

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
hrhino: You can do like:

```scala
def checkKeyLength[F[_]: ApplicativeError[?, Throwable]](kl: Int): F[Unit] =
  if (SS.crypto_kdf_BYTES_MIN > kl || kl > SS.crypto_kdf_BYTES_MAX)
    ApplicativeError[F].raiseError(kl)
  else ().pure[F]

def checkContextBytes = ...
// then...
val ctx = context.utf8Bytes
checkKeyLength(keyLength) *> checkContextBytes(ctx) *> F.delay {
  ...
}
```

jmcardon: I don't like the 3 flatmaps personally. Maybe im just a stickler for performance, but that's at least 3 calls into an `IO` data structure + flatmaps.

I am personally guilty of `throw` in tsec-libsodium. That said, it's because some of the conditionals are _after_ a side effect.

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/package.scala
jmcardon: :+1: on tsecerror

tsec-libsodium/src/test/scala/tsec/libsodium/KeyDerivationTest.scala
jmcardon: I have to agree here. These are the things we should be testing for re: key derivation.

tsec-libsodium/src/main/scala/tsec/libsodium/kdf/KeyDerivation.scala
jmcardon: Lol I deleted the wrong comment accidentally.

I was gonna say, even if this style is less functional, it's most certainly more efficient (no flatmaps, less allocations), so I'm ok with it.

PR #74, opened 2017-12-15
Initial One-time auth

rsoeldner:


tsec-libsodium/src/main/scala/tsec/libsodium/authentication/OnetimeAuth.scala
```diff
@@ -0,0 +1,48 @@
...
+    tag
+  }
+
+  def verify[F[_]](key: SodiumKey[OnetimeAuth], message: Array[Byte], tag: SodiumKey[MessageTag])(
+      implicit F: Sync[F],
+      S: ScalaSodium
+  ): F[Unit] = F.delay {
+
+    if (S.crypto_onetimeauth_verify(tag, message, message.length, key) != 0)
+      throw InvalidMessage
```
hrhino: I don't like this at all. Verification should return a `F[Boolean]`; it's not necessarily an error for the tag to be invalid.

tsec-libsodium/src/main/scala/tsec/libsodium/authentication/OnetimeAuth.scala
```diff
@@ -0,0 +1,48 @@
...
+import tsec.libsodium.ScalaSodium
+import tsec.libsodium.cipher.SodiumKey
+
+sealed trait OnetimeAuth
+
+sealed trait MessageTag
+
+object OnetimeAuth {
+
+  case object InvalidMessage extends Exception with Product with Serializable {
```
hrhino: `with Product with Serializable` is unneeded I think

rsoeldner: I think this style is used over the complete code base. Should be change this everywhere ?

tsec-libsodium/src/test/scala/tsec/libsodium/authentication/OnetimeAuthTest.scala
```diff
@@ -0,0 +1,38 @@
...
+      program.unsafeRunSync() mustBe trueg
+    }
+  }
+
+  it should "not verify manipulated message" in {
+    forAll { (s: String) =>
+      val program: IO[Boolean] = for {
+        key <- OnetimeAuth.generateKey[IO]
+        message  = s.utf8Bytes
+        message2 = ("$" + s).utf8Bytes
```
hrhino: I would do a `forAll { (s1, s2) => s1 == s2 || program.unsafeRunSync() mustBe false }`.

tsec-libsodium/src/main/scala/tsec/libsodium/authentication/OnetimeAuth.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def verify[F[_]](key: SodiumKey[OnetimeAuth], message: Array[Byte], tag: SodiumKey[MessageTag])(
+      implicit F: Sync[F],
+      S: ScalaSodium
+  ): F[Boolean] = F.delay {
+
+    if (S.crypto_onetimeauth_verify(tag, message, message.length, key) == 0)
+      true
+    else
+      false
```
hrhino: please no `if (foo) true else false`!

tsec-libsodium/src/main/scala/tsec/libsodium/authentication/OnetimeAuth.scala
hrhino: Maybe. @jmcardon was there a reason?

tsec-libsodium/src/test/scala/tsec/libsodium/authentication/OnetimeAuthTest.scala
rsoeldner: I used `whenever` now, thanks for the help :+1:

tsec-libsodium/src/main/scala/tsec/libsodium/authentication/OnetimeAuth.scala
```diff
@@ -0,0 +1,37 @@
...
+sealed trait OnetimeAuth
+
+sealed trait MessageTag
+
+object OnetimeAuth {
+
+  def generateKey[F[_]](implicit F: Sync[F], S: ScalaSodium): F[SodiumKey[OnetimeAuth]] = F.delay {
+    val key = SodiumKey[OnetimeAuth](new Array[Byte](ScalaSodium.crypto_onetimeauth_poly1305_KEYBYTES))
+
+    S.crypto_onetimeauth_keygen(key)
```
jmcardon: This is ok,  but just for the record, this is just `randombytes_buf` lol. I've avoided using the `x_keygen` functions for this reason.

rsoeldner: Ok if @hrhino and travis are ok too, i would sqash my commits to a single one.

PR #83, opened 2017-12-23
Add primitive convenience helper syntax, closes #66

### LINKED ISSUE #66: New convenience methods for next milestone
We work on strings and bytes a lot,  but some stuff we're missing:

Translations from primitives => byte => primitive for:
 * Long
 * Int
 * Double
 * Short
 * Float

(Anyone attempting to to bool => bytes needs to sincerely rethink their life).

rsoeldner:


common/src/main/scala/tsec/common/package.scala
```diff
@@ -17,13 +18,32 @@ package object common {
...
+
+  implicit final class TSecLongOps(val v: Long) extends AnyVal {
+    def toBytes: Array[Byte] = ByteBuffer.allocate(8).putLong(v).array()
+  }
+
+  implicit final class TSecIntOps(val v: Int) extends AnyVal {
+    def toBytes: Array[Byte] = ByteBuffer.allocate(4).putInt(v).array()
+  }
+
+  implicit final class TSecShortOps(val v: Short) extends AnyVal {
```
jmcardon: Could these be in the same style as the others? Aka final class, implicit def.

Simply for consistency's sake

common/src/main/scala/tsec/common/package.scala
```diff
@@ -17,13 +18,32 @@ package object common {
...
 
-
   trait StringNewt {
     type I <: String
 
     val is: Is[I, String]
   }
 
+  implicit final class TSecFloatOps(val v: Float) extends AnyVal {
+    def toBytes: Array[Byte] = ByteBuffer.allocate(4).putFloat(v).array()
```
jmcardon: Is there a source for these numbers? Seem like magic. A comment at the top of the syntax block is fine

rsoeldner: :+1:

common/src/main/scala/tsec/common/package.scala
```diff
@@ -17,12 +18,31 @@ package object common {
     override def getMessage: String = cause
   }
 
-
   trait StringNewt {
     type I <: String
 
     val is: Is[I, String]
   }
+  // ByteBuffer capacity based on https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html
```
ChristopherDavenport: Could this be mad into a typeclass with bitraversal to and from the type being encoded in its laws. This is quite lawless presently and doesn't have to be as this follows a standard codec pattern.

ChristopherDavenport: Or atleast as a `ByteEncoder` which imlies the same behavior. I believe this forms a `Prism` so prism laws could apply, but I may be wrong.

rsoeldner: @ChristopherDavenport Happy Christmas and thank you for your help ;-)

Issue #88, opened 2017-12-26
TSecAuthService composition via SemigroupK


jarreds:
We should be able to:

```scala
val serviceOne: TSecAuthService[IO, DummyUser, Authenticator[Int]] = ???
val serviceTwo: TSecAuthService[IO, DummyUser, Authenticator[Int]] = ???
serviceOne <+> serviceTwo
```

Currently, this results in a compiler error. I believe we have an issue similar to https://github.com/http4s/http4s/pull/1509

jmcardon:
I agree. Would you like to submit a PR? I am a bit gated in how much I can work from China until I am back home.

PR #90, opened 2017-12-26
Change type param order in TSecAuthService for partial unification

jarreds:
Similar to https://github.com/http4s/http4s/pull/1509

Resolves https://github.com/jmcardon/tsec/issues/88


tsec-http4s/src/main/scala/tsec/authentication/package.scala
```diff
@@ -63,23 +63,23 @@ package object authentication {
       }
   }
 
-  type TSecAuthService[F[_], Ident, A] = Kleisli[OptionT[F, ?], SecuredRequest[F, Ident, A], Response[F]]
+  type TSecAuthService[A, Ident, F[_]] = Kleisli[OptionT[F, ?], SecuredRequest[F, Ident, A], Response[F]]
```
rsoeldner: Please create a comment on the reason / issue here

tsec-http4s/src/test/scala/tsec/authentication/TSecAuthServiceSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+import tsec.mac.imports.HMACSHA256
+
+import cats.effect.IO
+import cats.implicits._
+
+class TSecAuthServiceSpec extends AuthenticatorSpec {
+
+  behavior of "Composability of TSecAuthService"
+
+    it should "compile when combined via SemigroupK" in {
```
jmcardon: Should compile when combined via semigroupK isn't exactly the condition we want. That's what this PR is about, but not what the test itself should reflect.

"Should apply left" or apply right ish tests to just confirm the behavior is working fine for semigroupK is what we want, because it also gives us a guarantee that somewhere up the pipeline we didnt do something funky

jmcardon:
Also @jarreds thank you very much for the contribution to the project!!!

I'd like for this to make it onto the next patch, its a much needed qol fix

tsec-http4s/src/test/scala/tsec/authentication/TSecAuthServiceSpec.scala
jmcardon: For an example, see: https://github.com/http4s/http4s/blob/215a84334b0a7b116f5b23f5feaf446843bf5054/server/src/test/scala/org/http4s/server/middleware/authentication/AuthMiddlewareSpec.scala#L149

jarreds: Agreed. I’ll get a new commit  up in the next hour or so.

PR #92, opened 2017-12-27
remove useless optionTs

jmcardon:
A lot of liftF was making the use of the tsec-http4s api force users to supply getOrElse for ops wherein OptionT wasn't the desired behavior

### DIFF
--- a/tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
+++ b/tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
@@ -223,7 +223,7 @@ object EncryptedCookieAuthenticator {
           raw: AEADCookie[A],
           now: Instant
       ): OptionT[F, AuthEncryptedCookie[A, I]] =
-        if (validateCookie(internal, raw, now)) refresh(internal) else OptionT.none
+        if (validateCookie(internal, raw, now)) OptionT.liftF(refresh(internal)) else OptionT.none
 
       def extractRawOption(request: Request[F]): Option[String] =
         unliftedCookieFromRequest(settings.cookieName, request).map(_.content)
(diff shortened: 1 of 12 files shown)


tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
```diff
@@ -454,20 +448,17 @@ object EncryptedCookieAuthenticator {
...
+      def refresh(authenticator: AuthEncryptedCookie[A, I]): F[AuthEncryptedCookie[A, I]] =
         settings.maxIdle match {
           case Some(_) =>
-            OptionT
-              .liftF(
-                F.delay(
-                  authenticator.copy[A, I](
-                    lastTouched = Some(Instant.now())
-                  )
+            F.delay(
```
rsoeldner: For consistency `F.delay(Instant.now()).flatMap( .. )`

tsec-http4s/src/main/scala/tsec/authentication/JWTAuthenticator.scala
```diff
@@ -569,7 +557,7 @@ object JWTAuthenticator {
 
       def parseRaw(raw: String, request: Request[F]): OptionT[F, SecuredRequest[F, V, AugmentedJWT[A, I]]] =
         for {
-          eInstance   <- OptionT.liftF(F.fromEither(enc.instance))
+          eInstance   <- OptionT.liftF(F.fromEither(enc.instance)) // Todo: pls my heart
```
rsoeldner: 😭

tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
jmcardon: Wait why? We're delaying the evaluation of `Instant.now` anyway, but we avoid one flatMap op. I think this is a small micro optimization.

PR #97, opened 2017-12-27
Use the Circe Java8 Instant decoder/encoder

jarreds:
Rather than a homespun one.

I was seeing implicit conflicts between tsec and circe.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,7 +66,8 @@ lazy val jwtCommonLibs = libraryDependencies ++= Seq(
   Libraries.circeCore,
   Libraries.circeGeneric,
   Libraries.circeGenericExtras,
-  Libraries.circeParser
+  Libraries.circeParser,
+  Libraries.circeJava8
 )
 
 lazy val http4sDeps = libraryDependencies ++= Seq(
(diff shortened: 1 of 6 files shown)


rsoeldner:
@jarreds TY! 

I had a quick look at [circe](https://github.com/circe/circe/blob/master/modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala#L27) and [java8 time](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#toString--) and I think it is in ISO-8601 format rather then using second since linux epoch. I'm unsure if this is the correct behavior. 

@jmcardon Please have a look here

jarreds:
Shoot, yeah, that's definitely not backward compatible. Thinking how we can prevent the conflict of having two encoders/decoders for `Instant` if you use circe java8 in an upstream project.

hrhino:
In which situations are you seeing conflicts?

Possibly we should simply make our codecs `private[tsec]`?

jarreds:
> In which situations are you seeing conflicts?

I had a couple `import tsec.authentication._` wildcard imports in an http4s service that needed to serialize an `Instant` in the ISO8601 format with the circe java8 encoder -- the two collided there. I can get around it with explicit imports so the tsec encoder isn't brought in scope eg:

```scala
import tsec.authentication.AuthEncryptedCookie
import tsec.authentication.TSecAuthService
import tsec.authentication.asAuthed
```

hrhino:
> `import tsec.authentication._` wildcard imports

Yeah. @jmcardon what do you think about making this `private[tsec]`?

jmcardon:
I'm ok with this being `private[tsec]`. Seems like a good WIP fix, but the true fix is getting rid of the `EncryptedCookie` internal garbage mess I made given that I don't need a whole json object for that crap. It doesn't give any extra security it being in a json object.

PR #101, opened 2017-12-29
Use tangly plugincat scalac options.

hrhino:
Allow for unused implicit params for now, since taking them out might cause binary compatibility concerns. I'm happy to go some other way on this one.

`-Ywarn-numeric-widen` is disabled on account of offense against the sundry `Constants` files, and I don't feel like busy-working them into compliance.

There are a lot of types defined in `package object`s as of this moment; I'll consider moving them in a later commit.

The entire `examples` module gets it disabled for lack of interest.


bench/src/main/scala/tsec/CipherBench.scala
```diff
@@ -55,19 +56,19 @@ class CipherBench {
...
   @Benchmark
-  def testLibSodiumAES(): Unit =
+  def testLibSodiumAES(): Any =
     AES256GCM
       .encrypt[IO](nPlaintext, sodiumAESKey)
       .unsafeRunSync()
 
   @Benchmark
-  def testLibSodiumXChacha20(): Unit =
+  def testLibSodiumXChacha20(): Any =
```
ChristopherDavenport: Why the `Any` - Value discard?  If so `.encrypt(...).void.unsafeRunSync` : `_=> ()` should be manageable latency

build.sbt
```diff
@@ -4,20 +4,8 @@ name := "tsec"
...
-  "utf8",
-  "-Ywarn-adapted-args", // Warn if an argument list is modified to match the receiver.
-  "-Ywarn-inaccessible", // Warn about inaccessible types in method signatures.
-  "-Ywarn-unused:imports",
-  "-Ywarn-nullary-override", // Warn when non-nullary overrides nullary, e.g. def foo() over def foo.
-  "-Ypartial-unification",
-  "-language:higherKinds",
-  "-language:implicitConversions"
-)
+// maybe enable ^C depending on the wind direction
```
ChristopherDavenport: The wind in this one has never been clear. Sometimes is nice though.

bench/src/main/scala/tsec/CipherBench.scala
hrhino: @ChristopherDavenport I would prefer not to introduce the noise of one more `map`; these benchmarks are going to be a little messy either way.

@jmcardon [if you return a value, it will `bh.consume` it](http://tutorials.jenkov.com/java-performance/jmh.html#return-value-from-benchmark-method).

rsoeldner: ha this is good to know :+1:

jmcardon:
I want to start using this now taht I finally have a breather. 

I'll pick out this PR and see what I can do. I think we have to start from scratch, as the repo has changed a lot since december, but it should be manageable.

hrhino:
👍

PR #109, opened 2017-12-30
Remove unused statement

rsoeldner:


jmcardon:
This is NOT an unused statement.

This checks whether the parsed as header id the correct algorithm. Otherwise, you are violating the spec.

This will introduce a bug in the code base.

PR #111, opened 2018-01-08
Cipher Rework

### LINKED ISSUE #99: Clean up tire fire around Symmetrics ciphers and DefaultEncryptor
JCA Symmetric Encryption is working and fine, but using it _sucks_.

I'll be straight since I was the one that wrote it: Using it as a user is _garbage_. Not because it's incorrect and not working, but because it's _too low level_. 

Here are my perceived griefs with symmetric ciphers:
- The FP version is not the default. What in the java was i doing...
- I made instance creation unsafe (Which is fair), which sort of makes users have to know the underlying cipher/mode/padding to even resolve it implicitly. The unsafe part isn't wrong, but it's an implementation detail that could be hidden better. 
- Users have to carry around _4 goddamn type params_ to use the functional version. Talk about annoying, seriously.
- In the JCA version, Plaintext is a case class, vs a newt.
- AAD is also a case class.
- AES-GCM has no detached mode for JCA. Storing the tag separately (or just getting a reference to it, why not) requires user JCA knowledge... that's pretty poor.
- Docs on key management sorely lacking.
- AES (especially useful for GCM mode of op) could have incremental nonce gen, which reduces the chance of a birthday attack for random nonce gen. 

Internal stuff that could be changed:
- If we generate the IVs manually, we don't ever need to call geIv 

It's a work in progress that impacts encryption in the long term .

jmcardon:
This pr is intended to improve and rework the ciphers module to address all points in #99 

Closes #99

Changes:
- Ciphers are now FP-first, and FP-only. No either interpreter bullcrap
- Instance creation is still wrapped in an effect (as it is unsafe), but because of the layout of the types and aliases, the user does not have to carry so many type params. It becomes much easier for a user to directly use our high level api.
- Direct reference to a low-level api, in the `primitives` package.
- Illegal combinations are now unrepresentable, due to the `IvProcess` typeclass. So no `DES/GCM/PKCS7Padding` will compile under tsec's types
- Users can now have more fine-grained tuning about IV generation or simply supplying it to the cipher. The `IvStrategy` type represents these types of constructions.

The following are a work in progress:
- [x] Plaintext is a case class, should be a newt
- [x] AAD is also a case class.
- [x] AES-GCM cipher tag detached mode.
- [x] AES CTR and GCM incremental nonce IV strategy


jmcardon:
removed review requests since it's not completed yet.

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+
+import tsec.cipher.symmetric.{AAD, AuthTag, CipherText, PlainText}
+
+trait CipherAlgebra[F[_], A, M, P, K[_]] {
+
+  /** Encrypt our plaintext with a tagged secret key
+    *
+    * @param plainText the plaintext to encrypt
+    * @param key       the SecretKey to use
+    * @return
```
rsoeldner: missing _iv_ parameter

rsoeldner: missing _iv_ parameter

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+}
+
+//Todo: Abstract over auth tag, combined and non-combined
+trait AEADAlgebra[F[_], A, M, P, K[_]] extends CipherAlgebra[F, A, M, P, K] {
+
+  /** Encrypt our plaintext with a tagged secret key
+    *
+    * @param plainText the plaintext to encrypt
+    * @param key       the SecretKey to use
+    * @return
```
rsoeldner: same, missing _iv_

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+    */
+  def encryptDetached(plainText: PlainText, key: K[A], iv: Iv[A, M]): F[(CipherText[A, M, P], AuthTag[A])]
+
+  /** Decrypt our ciphertext
+    *
+    * @param cipherText the plaintext to encrypt
+    * @param key        the SecretKey to use
+    * @return
+    */
```
rsoeldner: missing _tag_ parameter

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+  def decryptDetached(cipherText: CipherText[A, M, P], key: K[A], tag: AuthTag[A]): F[PlainText]
+
+  /** Encrypt our plaintext using additional authentication parameters,
+    * Primarily for GCM mode and CCM mode
+    * Other modes will return a cipherError attempting this
+    *
+    * @param plainText the plaintext to encrypt
+    * @param key       the SecretKey to use
+    * @param aad       The additional authentication information
+    * @return
```
rsoeldner: missing _iv_

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+  def decryptAAD(cipherText: CipherText[A, M, P], key: K[A], aad: AAD): F[PlainText]
+
+  /** Encrypt our plaintext using additional authentication parameters,
+    * Primarily for GCM mode and CCM mode
+    * Other modes will return a cipherError attempting this
+    *
+    * @param plainText the plaintext to encrypt
+    * @param key       the SecretKey to use
+    * @param aad       The additional authentication information
+    * @return
```
rsoeldner: missing _iv_

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/CipherAlgebra.scala
```diff
@@ -0,0 +1,88 @@
...
+    */
+  def encryptAADDetached(plainText: PlainText, key: K[A], iv: Iv[A, M], aad: AAD): F[(CipherText[A, M, P], AuthTag[A])]
+
+  /** Decrypt our ciphertext using additional authentication parameters,
+    * Primarily for GCM mode and CCM mode
+    * Other modes will return a cipherError attempting this
+    *
+    * @param cipherText the plaintext to encrypt
+    * @param key        the SecretKey to use
+    * @param aad        The additional authentication information
```
rsoeldner: missing _tag_

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/IvStrategy.scala
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+trait CounterIvStrategy[A, M] extends IvStrategy[A, M] {
+  def numGenerated[F[_]](implicit F: Sync[F]): F[Long]
+
+  def unsafeNumGenerated: Long
+}
+
+private[tsec] class FailedAtomicLong extends AtomicLong {
```
rsoeldner: why ?

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/imports/TripleDES.scala
```diff
@@ -0,0 +1,5 @@
+package tsec.cipher.symmetric.imports
+
+sealed trait TripleDES
+
+object TripleDES extends BlockCipherEV[TripleDES]("DESede", 8, 24)
```
rsoeldner: I didn't know `DESede` is a synonym for `TripleDES`. Maybe we should add a note here ?

cipher-symmetric/src/test/scala/tsec/SymmetricSpec.scala
```diff
@@ -78,71 +82,73 @@ class SymmetricSpec extends TestSpec with MustMatchers with PropertyChecks {
...
-          instance  <- JCAAEADImpure[A, M, P]
-          encrypted <- instance.encrypt(testPlainText, key1)
-          decrypted <- instance.decrypt(encrypted, key2)
-        } yield new String(decrypted.content, "UTF-8")
+        val testEncryptionDecryption: IO[String] = for {
+          key1      <- keyGen.generateLift[IO]
+          key2      <- keyGen.generateLift[IO]
+          encrypted <- algebra.encrypt[IO](testPlainText, key1)
+          decrypted <- algebra.decrypt[IO](encrypted, key2)
+        } yield new String(decrypted, "UTF-8")
```
rsoeldner: `.toUtf8String` instead of `new String(decrypted, "UTF-8")`

cipher-symmetric/src/test/scala/tsec/SymmetricSpec.scala
```diff
@@ -151,14 +157,82 @@ class SymmetricSpec extends TestSpec with MustMatchers with PropertyChecks {
...
-          key1      <- keyGen.generateKey()
-          instance  <- JCAAEADImpure[A, M, P]
-          encrypted <- instance.encryptAAD(testPlainText, key1, aad1)
-          decrypted <- instance.decryptAAD(encrypted, key1, aad2)
-        } yield new String(decrypted.content, "UTF-8")
+        val testEncryptionDecryption: IO[String] = for {
+          key1      <- keyGen.generateLift[IO]
+          encrypted <- algebra.encryptWithAAD[IO](testPlainText, key1, aad1)
+          decrypted <- algebra.decryptWithAAD[IO](encrypted, key1, aad2)
+        } yield new String(decrypted, "UTF-8")
```
rsoeldner: same here

cipher-symmetric/src/test/scala/tsec/SymmetricSpec.scala
```diff
@@ -151,14 +157,82 @@ class SymmetricSpec extends TestSpec with MustMatchers with PropertyChecks {
...
+
+    it should "not decrypt properly for an incorrect key in detached mode" in {
+      forAll { (testMessage: String) =>
+        val testPlainText = PlainText(testMessage.utf8Bytes)
+        val testEncryptionDecryption: IO[String] = for {
+          key1      <- keyGen.generateLift[IO]
+          key2      <- keyGen.generateLift[IO]
+          encrypted <- algebra.encryptDetached[IO](testPlainText, key1)
+          decrypted <- algebra.decryptDetached[IO](encrypted._1, key2, encrypted._2)
+        } yield new String(decrypted, "UTF-8")
```
rsoeldner: `.toUtf8String`

cipher-symmetric/src/test/scala/tsec/SymmetricSpec.scala
```diff
@@ -151,14 +157,82 @@ class SymmetricSpec extends TestSpec with MustMatchers with PropertyChecks {
...
+    it should "not decrypt properly for correct key but incorrect AAD in detached mode" in {
+      forAll { (testMessage: String, AAD1: String, AAD2: String) =>
+        val testPlainText = PlainText(testMessage.utf8Bytes)
+        val aad1          = AAD(AAD1.utf8Bytes)
+        val aad2          = AAD(AAD2.utf8Bytes)
+        val testEncryptionDecryption: IO[String] = for {
+          key1      <- keyGen.generateLift[IO]
+          encrypted <- algebra.encryptWithAADDetached[IO](testPlainText, key1, aad1)
+          decrypted <- algebra.decryptWithAADDetached[IO](encrypted._1, key1, aad2, encrypted._2)
+        } yield new String(decrypted, "UTF-8")
```
rsoeldner: `.toUtf8String`

common/src/main/scala/tsec/common/ByteUtils.scala
```diff
@@ -4,15 +4,54 @@ import java.security.MessageDigest
...
+      (a >> 16).toByte,
+      (a >> 8).toByte,
+      a.toByte
+    )
+
+  @inline def unsafeBytesToInt(a: Array[Byte]): Int =
+    a(0) << 24 | (a(1) & 0xFF) << 16 | (a(2) & 0xFF) << 8 | (a(3) & 0xFF)
+
+  @inline def bytesToInt(a: Array[Byte]): Option[Int] =
+    if (a.length < Integer.BYTES) None
```
rsoeldner: shouldn't this `a.length != Integer.BYTES`

common/src/main/scala/tsec/common/ByteUtils.scala
```diff
@@ -4,15 +4,54 @@ import java.security.MessageDigest
...
+      ((a(1).toLong & 0xff) << 48) |
+      ((a(2).toLong & 0xff) << 40) |
+      ((a(3).toLong & 0xff) << 32) |
+      ((a(4).toLong & 0xff) << 24) |
+      ((a(5).toLong & 0xff) << 16) |
+      ((a(6).toLong & 0xff) << 8) |
+      a(7).toLong & 0xff
+
+  @inline def bytesToLong(a: Array[Byte]): Option[Long] =
+    if (a.length < java.lang.Long.BYTES) None
```
rsoeldner: same here, I would prefer to match on equal size

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/core/IvStrategy.scala
jmcardon: because it was late at night and I didn't even notice I didn't delete that....

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/imports/TripleDES.scala
jmcardon: Eh, I'd actually do it backwards and leave TripleDES as the real name.

`DESede` means `DES encryption-decryption-encryption` which is a way to apply DES for encrypting things.

cipher-symmetric/src/test/scala/tsec/SymmetricSpec.scala
jmcardon: didn't notice, this is a leftover from writing a test before `toUtf8String` even existed.

common/src/main/scala/tsec/common/ByteUtils.scala
jmcardon: good question: Do we want to fail all byte arrays that are not the exact size, or should we be lenient on arrays that are at least N bytes?

cipher-symmetric/src/main/scala/tsec/cipher/symmetric/imports/TripleDES.scala
rsoeldner: thank you

common/src/main/scala/tsec/common/ByteUtils.scala
rsoeldner: In my opinion we should fail if the size doesn't match.

rsoeldner:
@jmcardon is this still wip ?

PR #112, opened 2018-01-10
Upgrading to Cats 1.0.1

gvolpe:
Simple upgrade to latest Cats. Just as a note, GitHub4s is still on Cats 1.0.0-RC2 and it won't upgrade until this issue is fixed: https://github.com/circe/circe/issues/819

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ lazy val scalacOpts = scalacOptions := Seq(
 )
 
 lazy val micrositeSettings = Seq(
-  libraryDependencies += "com.47deg" %% "github4s" % "0.17.0",
+  libraryDependencies += Libraries.gitHub4s,
   micrositeName := "TSec",
   micrositeBaseUrl := "/tsec",
   micrositeDescription := "A Type-Safe General Cryptography Library on the JVM",
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -4,18 +4,19 @@ import Keys._
 object Dependencies {
 
   object Versions {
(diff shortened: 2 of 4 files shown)


project/Dependencies.scala
```diff
@@ -4,18 +4,19 @@ import Keys._
 object Dependencies {
 
   object Versions {
-    val circeV        = "0.9.0-M3"
-    val catsV         = "1.0.0-RC2"
-    val catsEffV      = "0.6"
+    val circeV        = "0.9.0"
+    val catsV         = "1.0.1"
+    val catsEffV      = "0.7"
```
jmcardon: cats-effect 0.8 targets 1.0.1

gvolpe: Okay, didn't want to update it since both Http4s and Fs2 versions are based on Cats Effect 0.7, but I guess they are compatible so no harm.

jmcardon: If all tests pass it should be ok. I'm mostly trying to avoid the whole versioning issue

gvolpe:
Glad to do it :)

build.sbt
```diff
@@ -20,7 +20,7 @@ lazy val scalacOpts = scalacOptions := Seq(
 )
 
 lazy val micrositeSettings = Seq(
-  libraryDependencies += "com.47deg" %% "github4s" % "0.17.0",
+  libraryDependencies += Libraries.gitHub4s,
```
hrhino: 👍  to using `Dependencies`.

PR #113, opened 2018-01-31
Add Custom claims to JWTs

### LINKED ISSUE #59: Support JWT with non-standard headers
This is required for things like Google OAuth integration.

### LINKED ISSUE #76: Adding custom claims to JWTAuthenticator?
First, thanks for building this library. 

I'm using a `JWTAuthenticator` and it works great.  I want to add custom claims object to pass some information client side.  Since `AuthenticatorService.create` is implementation agnostic the best option I can think of is to make some sort of method which constructs JWT's a la https://jmcardon.github.io/tsec/docs/jwt-mac.html and avoid using `JWTAuthenticator.create`.  Is this your take as well? 

The docs could use a blurb on this, you open to contributors?

jmcardon:
Closes #76 
Closes #59 

Maybe the builder is redundant, I'm unsure.

### DIFF
--- a/common/src/main/scala/tsec/common/ByteUtils.scala
+++ b/common/src/main/scala/tsec/common/ByteUtils.scala
@@ -14,7 +14,7 @@ object ByteUtils {
     *  remove that crap
     *
     */
-  @inline final def find[A: Eq](arr: Array[A], p: A => Boolean): Boolean =
+  @inline final def find[A](arr: Array[A], p: A => Boolean): Boolean =
     if (arr.isEmpty)
       false
     else {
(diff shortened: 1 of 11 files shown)


common/src/main/scala/tsec/common/ByteUtils.scala
```diff
@@ -2,33 +2,75 @@ package tsec.common
...
 
-  @inline def zeroByteArray(a: Array[Byte]): Unit =
+  /** Only exists because the scala stdlib decided
+    * iterators were the best way to implement a .find operation.
+    *
+    *  Will remove iff the new collections
+    *  remove that crap
+    *
+    */
+  @inline final def find[A: Eq](arr: Array[A], p: A => Boolean): Boolean =
```
rsoeldner: I think we don't need to constrain A here.

jmcardon: True, will fix.

PR #115, opened 2018-02-09
Expire signed cookie on discard

aeons:
And streamline the expiry for encrypted cookies (by using `Instant.EPOCH`)

I looked at adding a test for it, but I couldn't figure out how to only
target the cookie-based authenticators.

### DIFF
--- a/tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
+++ b/tsec-http4s/src/main/scala/tsec/authentication/EncryptedCookieAuthenticator.scala
@@ -136,6 +136,13 @@ object AuthEncryptedCookie {
 
 object EncryptedCookieAuthenticator {
 
+  private[tsec] def isExpired[I: Encoder: Decoder, A: AES](
+      internal: AuthEncryptedCookie[A, I],
+      now: Instant,
+      maxIdle: Option[FiniteDuration]
+  ): Boolean =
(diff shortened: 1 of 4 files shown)


jmcardon:
@aeons Hi! Thanks for the PR.

Technically, the [`expireAuthenticator` test should cover this](https://github.com/jmcardon/tsec/blob/master/tsec-http4s/src/test/scala/tsec/authentication/RequestAuthenticatorSpec.scala#L82), but maybe here we need to create more specific tests.

aeons:
Yeah, I guess what I want is a specific test that the cookie is cleared on the client side by setting the expiry to some time in the past. I'm not sure how that would fit in in a test that also targets the token based authenticators.

jmcardon:
Hmm... This isn't exactly an easy problem.

I think part of the problem currently is just that in my tests, I (unfortunately) haven't made a distinction between failure to parse a token vs failure of the entire authentication in general.

I think this distinction needs to be done to successfully test this. 

Re: client-side, [it seems that this is the standard way to do this anyway](https://stackoverflow.com/questions/5285940/correct-way-to-delete-cookies-server-side).

aeons:
Yes, that's the implementation in this PR?

I agree that the test that "expired cookie === not authenticated" is there and works, what I would like to test is that `authenticator.discard => cookie is expired in response`.

But that won't make sense for a token based authenticator, so maybe adding a class of tests that only target the cookie based ones is the way to go.

Issue #122, opened 2018-03-01
Json decoding of JWTClaims fails when field "aud" is present


jfraudeau:
circe decoding that fails on
`aud <- c.downField(Audience).as[Option[Either[String, List[String]]]]`
 as the default decoder for Either expects a jsonObject with fields "Left" or "Right".
A workaround would be to use a custom decoder like mentionned in circe/circe#751

jmcardon:
Thank you, I'll put out a fix.

jmcardon:
Closed by #123

PR #123, opened 2018-03-01
 Add custom decoder for 'audience' in jwt

jfraudeau:
Solves https://github.com/jmcardon/tsec/issues/122

### DIFF
--- a/jwt-core/src/main/scala/tsec/jwt/JWTClaims.scala
+++ b/jwt-core/src/main/scala/tsec/jwt/JWTClaims.scala
@@ -38,7 +38,7 @@ import scala.util.control.NonFatal
 sealed abstract case class JWTClaims(
     issuer: Option[String],
     subject: Option[String],
-    audience: Option[Either[String, List[String]]], //case-sensitive
+    audience: Option[JWTAudience], //case-sensitive
     expiration: Option[Instant],
     notBefore: Option[Instant], // IEEE Std 1003.1, 2013 Edition time in seconds
     issuedAt: Option[Instant], // IEEE Std 1003.1, 2013 Edition time in seconds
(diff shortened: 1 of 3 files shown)


jwt-core/src/test/scala/tsec/jwt/JWTClaimsTest.scala
```diff
@@ -33,4 +33,18 @@ class JWTClaimsTest extends TestSpec {
     decode[JWTClaims](claims.asJson.toString()) mustBe Right(claims)
   }
 
+  it should "Deserialize the 'aud' field without failing" in {
```
jmcardon: Only missing a test for the list case, and we're good to go I think

jwt-core/src/main/scala/tsec/jwt/JWTClaims.scala
```diff
@@ -269,6 +269,16 @@ object JWTClaims extends JWSSerializer[JWTClaims] {
       }
   }
 
+  // Default Decoder[Either[A, B]] is not used (see: https://tools.ietf.org/html/rfc7519#section-4.1.3)
+  // Also this custom decoder "not" polymorphic as its behaviour could be inconsistent for other A and B
+  implicit val audienceDecoder: Decoder[Either[String, List[String]]] = {
+    c: HCursor => 
+      c.as[String] match {
+        case Right(a) => Right(Left(a))
```
jmcardon: This sort of making me not very sold on the signature. Maybe an `ADT` which you match on would be nice, but then we wouldn't get all of the stdlib goodies from `Either`

jfraudeau: An ADT seems apropriate here, I'm not sure how many people use the methods on Etiher from stdlib now but we could provide a `toList` method on the trait since in most case operations on the singleton list (e.g. finding the services own url in the list) will be equivalent to what would be done on the string literal (e.g. comparison).

PR #127, opened 2018-03-13
User NativePRNGNonBlocking for the ManagedRandom

### LINKED ISSUE #126: Use NativePRNGNonBlocking for ManagedRandom.
Currently the SecureRandom in ManagedRandom does not have a defined algorithm to use. We should use `NativePRNGNonBlocking`. It used /dev/urandom, which as pointed out [here](https://www.2uo.de/myths-about-urandom/) is as "random" as it gets.

coltfred:
Fixes #126

See also the conversation that was had around this: https://gitter.im/tsecc/Lobby?at=5aa1d22d6fba1a703ae55073


jmcardon:
Thank you for the contribution @coltfred!!!

PR #129, opened 2018-03-15
Add bouncy castle hash functions + Bouncy EV

### LINKED ISSUE #120: add Kekkack, Blake, Skein and Streebog family of Hashes on the jvm
From the bouncycastle provider, after #119 is merged.

jmcardon:
- Make bouncy castle require an evidence param that's loaded via implicits. It doesn't change the user-facing api AT ALL, but it _does_ cause the JVM to check whether the provder is loaded at least once :D 

Closes #120


bouncycastle/src/main/scala/tsec/Bouncy.scala
```diff
@@ -0,0 +1,16 @@
+package tsec
+
+import java.security.Security
+
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+
+trait Bouncy
```
rsoeldner: shouldn't this be _sealed_ ?

hashing-bouncy/src/main/scala/tsec/hashing/bouncy/hashes.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object Keccak384 extends AsBouncyCryptoHash[Keccak384]("SHA3-384")
+
+sealed trait Keccak512
+
+object Keccak512 extends AsBouncyCryptoHash[Keccak512]("SHA3-512")
+
+sealed trait Whirlpool
+
+object Whirlpool extends AsBouncyCryptoHash[Whirlpool]("Whirlpool")
```
rsoeldner: what a great name :D :tada:

hashing-bouncy/src/main/scala/tsec/hashing/bouncy/hashes.scala
```diff
@@ -0,0 +1,61 @@
...
+
+sealed trait Streebog256
+
+object Streebog256 extends AsBouncyCryptoHash[Streebog256]("GOST3411-2012-256")
+
+sealed trait Streebog512
+
+object Streebog512 extends AsBouncyCryptoHash[Streebog512]("GOST3411-2012-512")
+
+sealed trait Blake2B_160
```
rsoeldner: maybe its better to drop the `_`

bouncycastle/src/main/scala/tsec/Bouncy.scala
jmcardon: Can be, sure

hashing-bouncy/src/main/scala/tsec/hashing/bouncy/hashes.scala
jmcardon: What do you guys think?

@hrhino @ChristopherDavenport.

I added the underscore in this lonely case because reading `Blake2b160` seemed weird in my  mind. I don't think I have a single underscore anywhere else in the project.

Also, I would rather an underscore than ```Blake2b-160``` in backticks.

If you all hate it I will drop the underscores.

hrhino: Backticks are bad too IMO. Bouncy says `Blake2b160` and I think I like that a _little_ better. Maybe `Blake2B160` is a compromise.

All in all 🤷.

hashing-bouncy/src/main/scala/tsec/hashing/bouncy/package.scala
```diff
@@ -0,0 +1,45 @@
...
+import tsec.Bouncy
+import tsec.common.CryptoTag
+
+package object bouncy {
+
+  trait BouncyDigestTag[T] extends CryptoTag[T]
+
+  sealed class BouncyHasher[F[_], A](
+      implicit digestTag: BouncyDigestTag[A],
+      F: Applicative[F]
```
hrhino: 🙏

bouncycastle/src/main/scala/tsec/Bouncy.scala
hrhino: 👍 since it's a tag there's no reason to let ~~an attacker~~ some lazy fuck forge one

bouncycastle/src/main/scala/tsec/Bouncy.scala
```diff
@@ -0,0 +1,16 @@
...
+
+object Bouncy {
+  // @alexknvl
+  implicit val GetSchwifty: Bouncy = {
+    if (Security.getProvider("BC") == null)
+      Security.addProvider(new BouncyCastleProvider())
+    new Bouncy {}
+  }
```
hrhino: This is simultaneously disgusting and beautiful.

hashing-bouncy/src/main/scala/tsec/hashing/bouncy/hashes.scala
jmcardon: Alright it only takes at least two of you to convince me. `Blake2b160` it is.

bouncycastle/src/main/scala/tsec/Bouncy.scala
jmcardon: I agree, but I really don't see a better way around this :(.

It's between that, and having people's code fail at runtime like [we do right now on the docs](https://jmcardon.github.io/tsec/docs/signatures.html).

I'm personally pro-fixing this yolostyle approach.

jmcardon:
I'll merge this once #130 is clear :D

bouncycastle/src/main/scala/tsec/Bouncy.scala
hrhino: Yes I'm :+1: on it. Alternatively make the init happen as a side effect of the module ctor of everything that needs it, but that's fragile.

jmcardon:
Oh _shit_ I forgot to change the name of the blake fns.

PR #130, opened 2018-03-16
DJB Ciphers using bouncycastle

### LINKED ISSUE #124: DJB ciphers for JCA based off Bouncycastle.
`Salsa20`, `XSalsa20`, `Salsa20Poly1305`, so on and so forth are actually supported in bouncy castle (albeit they are inefficient as shit on the JVM, especially with a pure java impl compared to CPU compiler intrinsics as optimized by libsodium).

There's at least a small use case for people using this on the JVM that don';t wanna go the JNI route

jmcardon:
No way to make the PR smaller so here it goes:

This PR adds [XChacha20Poly1305IETF and Chacha20Poly1305 as well as XSalsa20Poly1305](https://download.libsodium.org/doc/secret-key_cryptography/aead.html) (Chacha20IETF I couldn't get working before I opened this PR so I will add it at a later date).

I'm not a fan of roll your own crypto unless absolutely necessary (which in this case, it is, there is no other implementation of these two that I find satisfactory. I do not like that google's tink use of bytebuffer). Hence, I was extra-thorough in this PR, as this adds:

- The usual tests (So a module specific test).
- Tests against libsodium using libsodium test vector + JNI instances
- Updated benchmarks for XChacha and XSalsa (named CryptoSecretBox in libsodium) on the jvm vs JNI (spoiler alert: JNI wins 10 fold).

Note: The implementation for XChacha20Poly1305 is a mix of what was done in `tink`, `BouncyCastle` and `libsodium`. 

Closes #124


cipher-bouncy/src/main/scala/tsec/cipher/symmetric/bouncy/ChaCha20Poly1305.scala
```diff
@@ -0,0 +1,270 @@
...
+    ): CipherText[ChaCha20Poly1305] =
+      unsafeEncryptAAD(plainText, k, iv, AAD(Array.empty[Byte]))
+
+    def unsafeEncryptAAD(
+        plainText: PlainText,
+        k: BouncySecretKey[ChaCha20Poly1305],
+        iv: Iv[ChaCha20Poly1305],
+        aad: AAD
+    ): CipherText[ChaCha20Poly1305] = {
+      val chacha20   = new ChaChaEngine(20)
```
rsoeldner: Can we add a link to a document describing all of these numbers ?

cipher-bouncy/src/main/scala/tsec/cipher/symmetric/bouncy/XSalsa20Poly1305.scala
```diff
@@ -0,0 +1,185 @@
...
+
+object XSalsa20Poly1305 extends AuthCipherAPI[XSalsa20Poly1305, BouncySecretKey] {
+  private val KeyLen        = 32
+  private val tagLen        = 16
+  private val NonceLenBytes = 24
+
+  implicit def defaultKeyGen[F[_]](implicit F: Sync[F]): SymmetricKeyGen[F, XSalsa20Poly1305, BouncySecretKey] =
+    new SymmetricKeyGen[F, XSalsa20Poly1305, BouncySecretKey] with ManagedRandom {
+      def generateKey: F[BouncySecretKey[XSalsa20Poly1305]] = F.delay {
+        val kBytes = new Array[Byte](KeyLen) //same as key len, 32 bytes
```
rsoeldner: :+1:

common/src/test/scala/tsec/SymmetricSpec.scala
```diff
@@ -0,0 +1,278 @@
...
+    it should "Encrypt and decrypt for the same key and AEAD in detached mode" in {
+      forAll { (testMessage: String, aadData: String) =>
+        val testPlainText = PlainText(testMessage.utf8Bytes)
+        val aad           = AAD(aadData.utf8Bytes)
+        val testEncryptionDecryption: IO[String] = for {
+          key       <- S.generateKey
+          encrypted <- E.encryptWithAADDetached(testPlainText, key, aad)
+          decrypted <- E.decryptWithAADDetached(encrypted._1, key, aad, encrypted._2)
+        } yield decrypted.toUtf8String
+        if (!testMessage.isEmpty)
```
rsoeldner: maybe we should use _whenever_

cipher-bouncy/src/main/scala/tsec/cipher/symmetric/bouncy/ChaCha20Poly1305.scala
jmcardon: I can yeah. Cryptographic code is pretty fucking full of magic numbers.

common/src/test/scala/tsec/SymmetricSpec.scala
jmcardon: tbh, I'm not a fan of using `scalatest` syntax, mostly because I don't want to be married to it.

To be quite honest, I do not enjoy scalatest in the slightest, nor specs2, but we live with it because there's nothing else (Yet at least. Edmund hyped me up a little bit ago about a testing library he's working on).

build.sbt
```diff
@@ -46,6 +46,7 @@ lazy val commonSettings = Seq(
   organization in ThisBuild := "io.github.jmcardon",
   scalaVersion in ThisBuild := "2.12.4",
   fork in test := true,
+  fork in run := true,
```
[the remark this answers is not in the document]
jmcardon: Yes, because I was testing stuff by using the `run` command and sbt doesn't release classloaders for native links, making multiple `run`s useless without this.

Hell, I think eventually when a PR comes for the `libsodium` module, i think that we should note to users that forking the jvm is the only way to run things in sbt that won't break and require a restart every two seconds :|. It's shitty, but try commenting it out, running some JNI code and seeing what happens. The results suck.

PR #142, opened 2018-03-22
QOL fixes for tsec-http4s

### LINKED ISSUE #104: Add a simpler non-verified parsing of JWTMac
In some use cases, we might simpl want to check whether the repr is fine, without redoing the signing work.

### LINKED ISSUE #105: JWTAuthenticators AugmentedJWT might do redundant work
to get an `AugmentedJWT` from the backend, we kinda have to parse the JWT portion of it.

This is redundant work for verification... We do all this work for every request submitted from a client, we don't need to even do a simpler reparse at all, we just need to check the serialized, raw version against the raw version in the server and call it a day.

### LINKED ISSUE #114: Make <+> AuthedService composition optional for non-spidering

jmcardon:
A few classes for authenticators seem to be over-constrained. 

This is in a general move to not force authenticators to use the JCA, to eventually enable fips for authentication as well.

This PR deals with:
- Loosening constrains, and removing unneeded ones.
- Adding an example similar to #140, for all authenticators
- Adds unverified parsing for a `JWTMac`. Small enough stuff to where it can fit in this PR too. 

Closes #105 
Closes #104 
Closes #114 

Note: #143 and #144 merge into this one as a general improvement of the Http4s api, to separate more significant points:
- #143 deals with general improvements to `PasswordStore`, from a usability persective
- #144 adds a user aware service.

Merge this one after #143 and #144


examples/src/main/scala/http4sExamples/BearerTokenExample.scala
```diff
@@ -54,4 +48,12 @@ object BearerTokenExample {
...
       Ok()
   }
+
+  val authedService2: AuthService = TSecAuthService {
+    case GET -> Root / "api2" asAuthed user =>
+      Ok()
+  }
+
+  val lifted: HttpService[IO] = Auth.liftService(authService1)
+  val liftedComposed: HttpService[IO] = Auth.liftService(authService1 <+> authedService2)
```
rsoeldner: :+1:

examples/src/main/scala/http4sExamples/jwtStatefulExample.scala
```diff
@@ -14,22 +15,24 @@ object jwtStatefulExample {
...
+    : MacSigningKey[HMACSHA256] = HMACSHA256.generateKey[Id] //Our signing key. Instantiate in a safe way using GenerateLift
 
   val jwtStatefulAuth =
     JWTAuthenticator.withBackingStore(
       expiryDuration = 10.minutes, //Absolute expiration time
-      maxIdle        = None,
-      tokenStore     = jwtStore,
-      identityStore  = userStore,
-      signingKey     = signingKey
+      maxIdle = None,
```
rsoeldner: what happens to your formating ?

jwt-mac/src/test/scala/tsec/JWTMacTests.scala
```diff
@@ -59,6 +60,24 @@ class JWTMacTests extends TestSpec with MustMatchers {
       res.unsafeRunSync() mustBe true
     }
 
+    it should "parse unverified correctly" in {
+      val res = for {
+        key    <- keyGen.generateKey
+        claims <- JWTClaims.withDuration[IO](notBefore = Some(-2.seconds), expiration = Some(10.seconds))
```
rsoeldner: Hm, I see negative seconds the first time, is this useful ?

tsec-http4s/src/main/scala/tsec/authentication/package.scala
```diff
@@ -79,12 +96,12 @@ package object authentication {
       * [[org.http4s.Response.notFound]], which generates a 404, for any request
       * where `pf` is not defined.
       */
-    def apply[A, I, F[_]](
+    def apply[I, A, F[_]](
```
rsoeldner: this change was because of partial unification bug ?

tsec-http4s/src/main/scala/tsec/authentication/package.scala
```diff
@@ -114,6 +131,62 @@ package object authentication {
       Kleisli.liftF(OptionT.none)
   }
 
+  type UserAwareService[Ident, A, F[_]] =
+    Kleisli[OptionT[F, ?], UserAwareRequest[F, Ident, A], Response[F]]
+
+  type UserAwareMiddleware[F[_], Ident, A] =
```
rsoeldner: `Ident` == `I` ?

tsec-http4s/src/main/scala/tsec/authentication/package.scala
jmcardon: yes.

jwt-mac/src/test/scala/tsec/JWTMacTests.scala
jmcardon: `notBefore` will just give you some `now.plusSeconds(t)` when the token can't be used.

This simply ensures after callign that function, `notBefore` always is valid.

examples/src/main/scala/http4sExamples/jwtStatefulExample.scala
jmcardon: idk ask scalafmt :(

tsec-http4s/src/main/scala/tsec/authentication/package.scala
jmcardon: yeah, same thing. 

`Ident` is any `Identity`

rsoeldner: for consistency we should change to `I`, what do you think ?

tsec-http4s/src/main/scala/tsec/authentication/package.scala
rsoeldner: :+1:

examples/src/main/scala/http4sExamples/jwtStatefulExample.scala
rsoeldner: haha :fire:

tsec-http4s/src/main/scala/tsec/authentication/package.scala
jmcardon: I don't disagree, sure. but `Ident` is also on `TSecAuthService`. Would you mind pushing to my branch real quick? I'm off to bed :)

tsec-http4s/src/main/scala/tsec/authentication/package.scala
jmcardon: Also to be clear: `A, I, F[_]` was changed because of the partial unification bug, then the swap to `I, A, F` was because the result type is `I, A, F`. It felt weird to hvae the type params in the apply method in different order of the result type.

tsec-http4s/src/main/scala/tsec/authentication/package.scala
rsoeldner: :+1:

PR #143, opened 2018-03-22
improve password store api

jmcardon:


tsec-http4s/src/main/scala/tsec/authentication/credentials/CredentialStore.scala
```diff
@@ -11,24 +12,76 @@ import tsec.passwordhashers.jca._
...
+  def putCredentials(raw: Array[Char], put: PasswordHash[P] => F[Unit]): F[Unit] =
+    for {
+      hash <- P.hashpw(raw)
+      _    <- put(hash)
+    } yield ()
+
+  def isAuthenticated(credentials: RawCredentials[Id]): F[Boolean] =
+    for {
+      pass <- retrievePass(credentials.identity)
+        .getOrElseF(F.raiseError(new NoSuchElementException("No such user")))
```
rsoeldner: Why not use https://github.com/jmcardon/tsec/blob/master/common/src/main/scala/tsec/common/package.scala#L15

tsec-http4s/src/main/scala/tsec/authentication/credentials/CredentialStore.scala
```diff
@@ -11,24 +12,76 @@ import tsec.passwordhashers.jca._
...
+    * @return
+    */
+  def putCredentials(credentials: C, put: P => F[Unit]): F[Unit]
 
-  def updateCredentials(credentials: C): F[Unit]
-
-  def removeCredentials(credentials: C): F[Unit]
+  /** Alias for `putCredentials`
+    * for when anyone may want to be verbose about
+    * put vs update
```
rsoeldner: :jack_o_lantern:

PR #149, opened 2018-03-29
Add a destructurable verification as an alternative to the booleans

jmcardon:
Edmund gave me a pretty good suggestion related to using a custom enum for a lot of our `F[Booleans]`

p.s Open to better name suggestions.

point being like

```scala
Blah.checkV[F](a, b).flatMap {
 case Verified => //..
 case VerificationFailed => //..
}
```

### DIFF
--- a/jwt-mac/src/main/scala/tsec/jws/mac/JWTMac.scala
+++ b/jwt-mac/src/main/scala/tsec/jws/mac/JWTMac.scala
@@ -93,7 +93,7 @@ object JWTMac {
       implicit hs: JWSSerializer[JWSMacHeader[A]],
       cv: JWSMacCV[F, A],
       F: Sync[F]
-  ): F[VerificationStatus] = F.delay(Instant.now()).flatMap(cv.verify(jwt.toEncodedString, key, _))
+  ): F[Boolean] = F.delay(Instant.now()).flatMap(cv.verifyBool(jwt.toEncodedString, key, _))
 
   def verifyFromString[F[_], A: JWTMacAlgo](jwt: String, key: MacSigningKey[A])(
       implicit s: JWSMacCV[F, A],


common/src/main/scala/tsec/common/VerificationStatus.scala
```diff
@@ -0,0 +1,5 @@
+package tsec.common
+
+sealed trait VerificationStatus extends Product with Serializable
+case object Verified           extends VerificationStatus
```
rsoeldner: I don't like the naming here, maybe this should be 
`VerificationOk` and `VerificationError` ?

jmcardon: `VerificationError` sounds like something other than not passing a check though, it's basically just a boolean that will force you to check semantically what it is. `VerificationError` sounds like it could be more than one thing, but really we have no such information.

hrhino:
What if we break everyone by making the "normal" ones return `ValidationStatus` and having `Blah.checkBool`?

common/src/main/scala/tsec/hashing/CryptoHasher.scala
```diff
@@ -19,4 +23,16 @@ trait CryptoHasher[F[_], A] {
     */
   def hashPipe: Pipe[F, Byte, Byte]
 
+  /** Check against another hash
+    *
+    */
+  def checkWithHash(l: Array[Byte], r: CryptoHash[A])(implicit F: Functor[F]): F[Boolean] =
+    F.map(hash(l))(MessageDigest.isEqual(_, r))
```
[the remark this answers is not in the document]
jmcardon: BUT AN EXTRA MAP OP :100: 

I'm not even kidding I did it this way to be pedantic about performance

PR #155, opened 2018-03-31
Update readme version and docs.

jmcardon:


jmcardon:
Probably a good opportunity to fix that tut travis failure

PR #170, opened 2018-05-25
update deps, fix deprecations

jmcardon:


tsec-http4s/src/main/scala/tsec/authentication/package.scala
```diff
@@ -93,6 +93,9 @@ package object authentication {
 
   object TSecAuthService {
 
+    //Lol hack but this works
+    private[this] val cachedUnauthorized: Response[Any] = Response[Any](Status.Unauthorized)
```
ChristopherDavenport: 👎

jmcardon:
Removed just for you @ChristopherDavenport :wink:

jmcardon:
For the record, I'm merging #171 and #173 into this one since I started working on them before this could be merged. We can merge them all as a batch in one pop.

PR #172, opened 2018-05-25
Update Scala Versions and Compiler Plugins

ChristopherDavenport:


jmcardon:
Thanks a lot @ChristopherDavenport

PR #173, opened 2018-05-25
(WIP) Key rotation for tsec-http4s

jmcardon:
This is a fairly simple but powerful implementation of key rotation based off of our primitives.

This is source compatible,but adds and changes quite a few sigs.

In particular, I would love if you guys could take a look at `TimedRotator`. I'm not sure if something is off there: I am relying on `Timer[F]` and `ExecutionContext` as well as `ConcurrentEffect`. I'm very unsure whether those should abstract and defined @ the impl site, without burdening the class with the heavy constraining. Opinions highly sought


jmcardon:
Pls ignore tut failures. Those are handled by #171 and I wouldn't merge this before that one

PR #175, opened 2018-06-15
add oauth2 support

Qi77Qi:


oauth2/src/main/scala/tsec.oauth2/provider/AuthorizationRequest.scala
```diff
@@ -0,0 +1,54 @@
...
+          case _ =>
+            clientCredentialByParam.toRight(InvalidClient(s"Failed to parse client credential from header (${e.description}) and params"))
+        }
+      },
+      Right(_)
+    )
+  }
+
+  private def clientCredentialByAuthorization(s: String): Either[InvalidClient, ClientCredential] =
+    Try(new String(Base64.getDecoder.decode(s), "UTF-8"))
```
jmcardon: 2 quips:

- [Tsec-common has a `.base64Bytes` helper, for the record.](https://github.com/jmcardon/tsec/blob/master/common/src/main/scala/tsec/common/package.scala#L54), as well as a utf8 one. This could be done by `s.base64Bytes.toUtf8String`. Granted, this is not safe and in a later PR I need to address this.
- I'm _very_ iffy on `Try` but I might be able to admit this could be a good use case for it.

oauth2/src/main/scala/tsec.oauth2/provider/DataHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+ * Access token
+ *
+ * @param token Access token is used to authentication.
+ * @param refreshToken Refresh token is used to re-issue access token.
+ * @param scope Inform the client of the scope of the access token issued.
+ * @param life Life of the access token since its creation.
+ * @param createdAt Access token is created date.
+ * @param params Additional parameters to add information/restriction on given Access token.
+ */
+final case class AccessToken(token: String, refreshToken: Option[String], scope: Option[String], life: Option[FiniteDuration], createdAt: Date, params: Map[String, String] = Map.empty[String, String]) {
```
jmcardon: why `Date` and not `Instant`? The latter is what people use nowadays. Not like we're worried about supporting jdk7 and below: we don't even build for it.

jmcardon: Also I'd prefer `lifeTime` over `life` here.

oauth2/src/main/scala/tsec.oauth2/provider/DataHandler.scala
```diff
@@ -0,0 +1,40 @@
...
+  def isExpired: Boolean = expiresIn.exists(_.toMillis < 0)
+
+  val expiresIn: Option[FiniteDuration] = life map { l =>
+    val expTime = createdAt.getTime + l.toMillis
+    ((expTime - System.currentTimeMillis) / 1000) milli
```
jmcardon: This thing will break RT _hard_ if this is evaluated wrong.

Anything that checks the system time is an effect. at _least_ `isExpired` should check the time every time, whether you choose to do it in an effect or not.

However, I do believe _isExpired_ should be `F[Boolean]` with `implicit F: Sync[F]` to account for the system check.

oauth2/src/main/scala/tsec.oauth2/provider/GrantHandler.scala
```diff
@@ -0,0 +1,166 @@
...
+final case class GrantHandlerResult[U](
+                                  authInfo: AuthInfo[U],
+                                  tokenType: String,
+                                  accessToken: String,
+                                  expiresIn: Option[FiniteDuration],
+                                  refreshToken: Option[String],
+                                  scope: Option[String],
+                                  params: Map[String, String])
+
+sealed trait GrantHandler {
```
jmcardon: `GrantHandler[F[_]]`

oauth2/src/main/scala/tsec.oauth2/provider/GrantHandler.scala
```diff
@@ -0,0 +1,166 @@
...
+    accessToken.scope,
+    accessToken.params)
+
+}
+
+object GrantHandler{
+
+  case object RefreshToken extends GrantHandler {
+    def handleRequest[U](request: AuthorizationRequest, handler: AuthorizationHandler[U]): IO[GrantHandlerResult[U]] = {
+      val res: EitherT[IO, Throwable, GrantHandlerResult[U]] = for{
```
jmcardon: `EitherT[IO, Throwable]` is completely redundant, as `IO` is `MonadError[F, Throwable]`.

I agree with the use case for `EitherT` for custom error lifting. For this, just stay in `IO`.

jmcardon: Same applies to the ones below this, btw.

oauth2/src/main/scala/tsec.oauth2/provider/ProtectedResourceHandler.scala
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Provide access to <b>Protected Resource</b> phase support for using OAuth 2.0.
+ *
+ * <h3>[Access to Protected Resource phase]</h3>
+ * <ul>
+ *   <li>findAccessToken(token)</li>
+ *   <li>findAuthInfoByAccessToken(token)</li>
+ * </ul>
+ */
+trait ProtectedResourceHandler[U] {
```
jmcardon: `F[_]`

oauth2/src/main/scala/tsec.oauth2/provider/package.scala
```diff
@@ -0,0 +1,16 @@
+package tsec.oauth2
+
+package object provider {
+  // grant_type
+  val AUTHORIZATION_CODE = "authorization_code"
```
jmcardon: These could form type-safe coproducts.

Not a fan of strings.

oauth2/src/test/scala/tsec.oauth2/provider/ImplicitSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+class ImplicitSpec extends FlatSpec with ScalaFutures with OptionValues {
+
+  it should "grant access with valid user authentication" in handlesRequest("user", "pass", true)
+  it should "not grant access with invalid user authentication" in handlesRequest("user", "wrong_pass", false)
+
+  def handlesRequest(user: String, pass: String, ok: Boolean) = {
+    val dataHandler = new ImplicitHandler[IO, MockUser]{
+      override def validateClient(maybeClientCredential: ClientCredential, request: ValidatedRequest): IO[Boolean] = IO.pure(true)
+
+      //        override def findUser(
```
jmcardon: Clean up?

PR #176, opened 2018-06-20
[WIP] JSON Web Key Set (JWKS) Support

sebastianvoss:


tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+
+final case class AugmentedJWKS[A, I](
+    id: SecureRandomId,
+    jwt: JWTSig[A],
+    identity: I,
+    expiry: Instant,
+    lastTouched: Option[Instant]
+)
+
+final case class JWK(kid: String, alg: String, kty: String, use: String, n: String, e: String)
```
jmcardon: Notes:
- JWK are simply a key representation format. Your "JWK Authenticator", as it is written, is a "JWK using an RSA public key authenticator". I'm ok with adding a specific use case (since you wrote it), but given this is definitely not generic, we either make it support the JWK spec for other things by making JWKs more generic, or we change this to state its actual intent
- This only seems to support RSA. Why?
- given this is only for RSA, why is `alg` not a fixed field?
- `n` and `e`  are public key parameters for RSA. This does not support RSA private keys.

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+
+final case class JWKS(keys: List[JWK])
+
+class KeyRegistry[F[_], A: JCASigTag](uri: Uri, minFetchDelay: FiniteDuration)(implicit E: Effect[F]) {
+
+  import io.circe.generic.auto._
```
jmcardon: Why are imports _inside_ the class? Are they conflicting? This is not a style this repository is using

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  private def shouldFetch() =
+    lastFetch.forall(_.isBefore(LocalDateTime.now().minusSeconds(minFetchDelay.toSeconds)))
+
+  private def getKey(id: String) = E.delay {
+    keys.get(id)
+  }
+
+  private def convert(jwk: JWK): SigPublicKey[A] = {
+    val modulus = new BigInteger(1, Base64.getUrlDecoder.decode(jwk.n))
```
jmcardon: We have syntax helpers for this in `tsec.common`

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
jmcardon: Moreover looking down, it seems you're using n and e as b64 encoded strings. I'd prefer newtypes for this if possible.

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+  implicit val jwksDecoder: EntityDecoder[F, JWKS] = jsonOf[F, JWKS]
+
+  private var keys = Map[String, SigPublicKey[A]]()
+  private var lastFetch: Option[LocalDateTime] = None
+
+  def getPublicKey(id: String): F[Option[SigPublicKey[A]]] = {
+    getKey(id).map {
+      case None if shouldFetch() =>
+        for {
+          _      <- E.delay(println("FETCH"))
```
jmcardon: either use a logger here, or you just forgot to remove a debug statement, lol

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  import io.circe.generic.auto._
+  import org.http4s.circe._
+  import org.http4s.client.blaze._
+  import java.security.KeyFactory
+  import java.security.spec.RSAPublicKeySpec
+
+  implicit val jwksDecoder: EntityDecoder[F, JWKS] = jsonOf[F, JWKS]
+
+  private var keys = Map[String, SigPublicKey[A]]()
```
jmcardon: This is making me feel like this could be accessed in a multithreaded scenario.

Use a `ref` or an `atomicReference` pls.

tsec-http4s/src/main/scala/tsec/authentication/JWKSAuthenticator.scala
```diff
@@ -0,0 +1,185 @@
...
+  private var lastFetch: Option[LocalDateTime] = None
```
jmcardon: ditto ref

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+
+final case class JWKS(keys: List[JWK])
+
+class KeyRegistry[F[_], A: JCASigTag](uri: Uri, minFetchDelay: FiniteDuration)(implicit E: Effect[F]) {
+
+  implicit val jwksDecoder: EntityDecoder[F, JWKS] = jsonOf[F, JWKS]
+  implicit val modulusDecoder: Decoder[Modulus]    = Decoder.decodeString.map(s => Modulus(BigInt(1, s.base64UrlBytes)))
+  implicit val exponentDecoder: Decoder[Exponent]  = Decoder.decodeString.map(s => Exponent(BigInt(1, s.base64UrlBytes)))
+
+  private val keys      = new AtomicReference(Map[String, SigPublicKey[A]]())
```
jmcardon: Okay, so the proper, functional way to do this, would be to make `keyRegistry` take an `Ref[F, Map[String, SigPublicKey[A]]` and `lastFetch` a `Ref[F, Instant]`.

For the record, `lastFetch` is _never_ set to `none` after being retrieved for the first time, so there's no reason to have it in option at all. Simply making the first set time something like the beginning of unix epoch ensures that the first call will always fetch.

Instantiation of `KeyRegistry` as a class is then `F[KeyRegistry[F, A]]`, which is fine and totally normal to signal the fact that it allocates mutable references

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+    getKey(id).map {
+      case None if shouldFetch() =>
+        for {
+          jwks      <- fetchJwks(uri)
+          k         <- E.delay(jwks.keys.map(jwk => (jwk.kid, convert(jwk))).toMap)
+          _         <- E.delay(keys.set(k))
+          _         <- E.delay(lastFetch.set(Instant.now().some))
+          key       <- E.delay(k.get(id))
+        } yield key
+      case _ => E.delay(keys.get().get(id))
```
jmcardon: This can be simplified to `case r => E.pure(r))`. You're calling `get()` twice for no reason whatsoever.

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+    }.flatten
+  }
+
+  private def fetchJwks(uri:Uri) = {
+    val s = for {
+      client <- Http1Client.stream[F]()
```
jmcardon: This creates an Http1Client _every single time_. 

Make the class take a `Client[F]`, otherwise this will be _excruciatingly slow_, as it will allocate a connection pool and then close it after you make your request. It makes no sense to allocate and then shutdown a connection pool for a single request every time.

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+
+  private def shouldFetch() =
+    lastFetch.get().forall(_.isBefore(Instant.now().minusSeconds(minFetchDelay.toSeconds)))
+
+  private def getKey(id: String) = E.delay {
+    keys.get().get(id)
+  }
+
+  private def convert(jwk: JWK): SigPublicKey[A] = {
+    val pubKey = KeyFactory.getInstance(jwk.kty).generatePublic(new RSAPublicKeySpec(jwk.n.value.bigInteger, jwk.e.value.bigInteger))
```
jmcardon: If this is an RSA key, `SigpublicKey[A]`, it worries me that you're accepting arbitrarily valued n and e parameters then parametrizing over anything SigPublicKey is. RSA in particular is subject to a lot of attacks for shitty values

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+    SigPublicKey[A](pubKey)
+  }
+
+}
+
+class JWKPublicKeyRSAAuthenticator[F[_] : Effect, I: Decoder, V, A: JCASigTag](
```
jmcardon: A: JCASigTag needs to be a stronger constraint or none at all in the class. JcaSigTag is for all sigs, not just RSA.

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -0,0 +1,197 @@
...
+    */
+  override def parseRaw(raw: String, request: Request[F]): OptionT[F, SecuredRequest[F, V, AugmentedJWK[A, I]]] =
+    OptionT(
+      (for {
+        now          <- Sync[F].delay(Instant.now())
+        extractedRaw <- cv.extractRaw(raw)
+        publicKey    <- keyRegistry.getPublicKey(extractedRaw.header.kid.get)
+        extracted    <- cv.verify(raw, publicKey.get, now)
+        jwtid        <- cataOption(extracted.body.subject)
+        id           <- cataOption(extracted.body.subject.flatMap(s => decode[I](s""""$s"""").toOption))
```
jmcardon: Why are you doing this? you can just write `decode[I](_).toOption`

sebastianvoss: When `I` is `String` the value passed to `decode` needs to be quoted. Otherwise Circe expects a json value:

```
scala> import io.circe.parser.decode
import io.circe.parser.decode

scala> decode[String]("user1")
res1: Either[io.circe.Error,String] = Left(io.circe.ParsingFailure: expected json value got u (line 1, column 1))

scala> decode[String](s""""user1"""")
res2: Either[io.circe.Error,String] = Right(user1)
```

jmcardon: yes, this is for a _string_ value, not a json object.

sebastianvoss: Wouldn't this also cause problems in e.g. [PartialStatelessJWTAuth](https://github.com/jmcardon/tsec/blob/194c26f3b45a1d5292faee5bea7fd4f496713fcf/tsec-http4s/src/main/scala/tsec/authentication/internal/PartialStatelessJWTAuth.scala#L40)?

If I have an `IdentityStore[F[_], I, V]` and `I` is `String` the decoding `extracted.body.subject.flatMap(decode[I](_).toOption)` would always fail, no?

jmcardon: I think I'd be more concerned with someone having their identity type be a string in the first place.

sebastianvoss: If I have a value class representing my user identity and using `deriveUnwrappedDecoder` it would also fail, wouldn't it?

jmcardon: if it's correctly supplied by the user, the string itself will come with quotes. We should not be adding this ourselves.

sebastianvoss: Are you saying the `sub` field in the JWT should have quotes (e.g. `"sub": "\"user1\"",`)? Sorry for being a pain in the arse. Just want to understand how it is intended to work.

sebastianvoss: Also, if you are using something like auth0 you don't have much influence on this.

sebastianvoss: I changed this as you suggested. But still interested how to make tsec work for my use case where the `sub` is not in my control.

tsec-http4s/src/main/scala/tsec/authentication/JWKPublicKeyRSAAuthenticator.scala
```diff
@@ -43,17 +43,22 @@ class KeyRegistry[F[_], A: JCASigTag: RSAKFTag](
    client: Client[F]
 )(implicit E: Effect[F]) {
 
-  implicit val jwksDecoder: EntityDecoder[F, JWKS] = jsonOf[F, JWKS]
-  implicit val modulusDecoder: Decoder[Modulus]    = Decoder.decodeString.map(s => Modulus(BigInt(1, s.base64UrlBytes)))
-  implicit val exponentDecoder: Decoder[Exponent]  = Decoder.decodeString.map(s => Exponent(BigInt(1, s.base64UrlBytes)))
+  implicit val jwksDecoder: EntityDecoder[F, JWKS[A]] = jsonOf[F, JWKS[A]]
+  implicit val modulusDecoder: Decoder[Modulus]       = Decoder.decodeString.map(s => Modulus(BigInt(1, s.base64UrlBytes)))
+  implicit val exponentDecoder: Decoder[Exponent]     = Decoder.decodeString.map(s => Exponent(BigInt(1, s.base64UrlBytes)))
+  implicit val x5cDecoder: Decoder[SigPublicKey[A]]   = Decoder.decodeString.map(s => {
```
sebastianvoss: @jmcardon would this be a better way to get the public key? Does it make sense to additionally compare `n` and `e` from the JWK with the values of the generated cert?

PR #187, opened 2018-07-30
add OS agnostic secure random

jmcardon:
- Add OSUtil, since we might need this later anyway for embedding JNI
- Add a check for windows. Otherwise, we default to /urandom

### DIFF
--- a/common/src/main/scala/tsec/common/ManagedRandom.scala
+++ b/common/src/main/scala/tsec/common/ManagedRandom.scala
@@ -1,46 +1,27 @@
 package tsec.common
 
 import java.security.SecureRandom
-import java.util.concurrent.atomic.LongAdder
 
 /** A trait that manages a secureRandom instance.
   */
 trait ManagedRandom {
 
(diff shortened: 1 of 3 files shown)


jmcardon:
@rhyskeepence Please test this branch whenever you can on a windows machine.

pnosko:
Hi, I ran into the problem solved by this commit, but alas I can't see the code supporting windows in the RC1 version. I checked out the secureRand branch, built it and it works. When can I expect to see it working for windows please? Thank you!

jmcardon:
@pnosko hi, not to be rude, but I've made it very clear tsec does not support windows.

@ the time of this PR I was considering it, but any changes that windows might need, unless one maintainer actually uses windows, can't be tested since, well, none of us use it.

Appveyor is only enough to tell us what's wrong, not to debug all problems. 

I fully encourage a fork of tsec to work on windows that you can publish locally and/or feed in with a JAR, but until one of the maintainers decides to step up and help support it (and they don't actually have to, I mean, I'm clearly not), it won't happen.

PR #196, opened 2018-08-29
Fix misspelled method names in RSAKFTag and ECKFTag

fiadliel:
unsafegenerateKeyPairStrong => unsafeGenerateKeyPairStrong
unsafebuildPrivateKeyFromPoint => unsafeBuildPrivateKeyFromPoint

There are no users of these methods visible through GH search.


fiadliel:
I can't push, somebody with write access will have to do that for me. Thanks!

PR #200, opened 2018-09-28
Cats community upgrades

jarreds:
A big question I have on this is if we should rename the `Tsec*Service` to `Tsec*Routes` so we're lined up with http4s. I was thinking we could follow up with that.


jarreds:
On second glance, this is definitely still hard to read w/ whitespace only. Should wait on #199 and a rebase to do a legit review.

tsec-http4s/src/main/scala/tsec/authentication/SecuredRequestHandler.scala
```diff
@@ -9,20 +10,20 @@ import tsec.authorization._
 
 sealed abstract class SecuredRequestHandler[F[_], Identity, User, Auth](
     val authenticator: Authenticator[F, Identity, User, Auth]
-)(implicit F: MonadError[F, Throwable]) {
+)(implicit F: MonadError[F, Throwable], AE: ApplicativeError[Kleisli[OptionT[F, ?], Request[F], ?], Throwable]) {
```
jarreds: Implicit needed here because they changed something in cats so the the `ApplicativeError[F, Unit]` takes highest priority 👎

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
```diff
@@ -102,7 +103,9 @@ abstract class AuthenticatorSpec extends TestSpec with MustMatchers with Propert
     def dAll(): Unit = storageMap.clear()
   }
 
-  def AuthenticatorTest[A](title: String, authSpec: AuthSpecTester[A]) = {
+  def AuthenticatorTest[A](title: String, authSpec: AuthSpecTester[A])(
+      implicit AE: ApplicativeError[OptionT[IO, ?], Throwable]
```
jarreds: Same reason here.

goedelsoup:
this looks roughly equivalent to the changes i had active. that said, i'd also like to see this diff sans whitespace changes.

build.sbt
```diff
@@ -115,14 +115,14 @@ lazy val commonSettings = Seq(
...
-  scalaVersion in ThisBuild := "2.12.6",
+  crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.7"),
+  scalaVersion in ThisBuild := "2.12.7",
   fork in test := true,
   fork in run := true,
   parallelExecution in test := false,
   addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.7"),
   addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.2.4"),
-  version in ThisBuild := "0.0.1-M11",
+  version in ThisBuild := "0.1.0-M1",
```
goedelsoup: 👍

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
```diff
@@ -113,7 +116,7 @@ abstract class AuthenticatorSpec extends TestSpec with MustMatchers with Propert
           fromRequest <- authSpec.auth.extractAndValidate(authSpec.embedInRequest(Request[IO](), auth))
           _           <- OptionT.liftF(authSpec.dummyStore.delete(dummy1.id))
         } yield fromRequest)
-          .handleErrorWith(_ => OptionT.none)
+          .handleErrorWith((_: Throwable) => OptionT.none)
```
goedelsoup: maybe use `scala.Function.const` instead

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
```diff
@@ -149,7 +152,7 @@ abstract class AuthenticatorSpec extends TestSpec with MustMatchers with Propert
           req2     <- authSpec.auth.extractAndValidate(authSpec.embedInRequest(Request[IO](), renewed))
           _        <- OptionT.liftF(authSpec.dummyStore.delete(dummy1.id))
         } yield req2)
-          .handleErrorWith(_ => OptionT.none)
+          .handleErrorWith((_: Throwable) => OptionT.none)
```
goedelsoup: maybe use `scala.Function.const` instead

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
```diff
@@ -180,7 +183,7 @@ abstract class AuthenticatorSpec extends TestSpec with MustMatchers with Propert
           req2      <- authSpec.auth.extractAndValidate(authSpec.embedInRequest(Request[IO](), discarded))
           _         <- OptionT.liftF(authSpec.dummyStore.delete(dummy1.id))
         } yield req2)
-          .handleErrorWith(_ => OptionT.none)
+          .handleErrorWith((_: Throwable) => OptionT.none)
```
goedelsoup: maybe use `scala.Function.const` instead

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
```diff
@@ -216,7 +219,7 @@ abstract class AuthenticatorSpec extends TestSpec with MustMatchers with Propert
             req2     <- authSpec.auth.extractAndValidate(authSpec.embedInRequest(Request[IO](), renewed))
             _        <- OptionT.liftF(authSpec.dummyStore.delete(dummy1.id))
           } yield req2)
-            .handleErrorWith(_ => OptionT.none)
+            .handleErrorWith((_: Throwable) => OptionT.none)
```
goedelsoup: four later, noting these are obviously just nits: maybe use `scala.Function.const` instead

build.sbt
goedelsoup: just realized this is defined twice. can we eliminate that?

tsec-http4s/src/main/scala/tsec/authentication/SecuredRequestHandler.scala
jmcardon: Hm. 

Well, I made it `OptionT`, at the time, because our `HttpRoutes` were `OptionT`, and it just seemed to make sense.

To be honest, a real solution is:
- Enumerate the errors into an `AuthError` adt (that extends `NoStackTrace` and `Exception`
- `.recover` on specific sections of the error composition and re-raise (for efficiency)
- return `EitherT` by recovering over the whole stmt.

It's a possible solution, anyway.

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
jmcardon: Is this really not implicitly offered :/

tsec-http4s/src/main/scala/tsec/authentication/SecuredRequestHandler.scala
jarreds: I think the argument that `OptionT` is correct because it is the same type as `HttpRoutes` is still valid.

Made a quick change to the code to make why I added this more clear. Here is what I posted in the http4s channel to see if this was intentional, because cats had a breaking change:

> doing a port of tsec to 0.19 and noticed a change that (imo) reduces some ergonomics.
> we used to be able to do (svc: HttpService[F]).handleErrorWith(t: Throwable => ...)
> so i expected (routes: HttpRoutes[F]).handleErrorWith(t: Throwable => ...) would work the same
> but it seems the provided implicit expects a Unit
> eg (routes: HttpRoutes[F]).handleErrorWith(u: Unit => ...)
> i think this was because of a cats change?
> anyway, i think i can get around this with an implicit ME: MonadError[Kleisli[OptionT[F, ?], Request[F], ?], Throwable]
> and then ME.handleErrorWith(routes: HttpRoutes[F])(t: Throwable => ...)
> but that feels kinda meh.

I'm ok, leaving that as-is. What do you think?

tsec-http4s/src/test/scala/tsec/authentication/AuthenticatorSpec.scala
jarreds: Yeah, it's sad.

build.sbt
```diff
@@ -115,14 +115,13 @@ lazy val commonSettings = Seq(
...
-  crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.6"),
-  scalaVersion in ThisBuild := "2.12.6",
+  crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.7"),
+  scalaVersion in ThisBuild := "2.12.7",
   fork in test := true,
   fork in run := true,
   parallelExecution in test := false,
   addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.7"),
   addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.2.4"),
-  version in ThisBuild := "0.0.1-M11",
```
jarreds: Version now defined in `version.sbt` exclusively.

tsec-http4s/src/main/scala/tsec/authentication/SecuredRequestHandler.scala
jarreds: Also going to ask if they can put this implicit in http4s so we won't need it here.

jarreds:
Any other changes we want in this PR?

PR #207, opened 2019-02-06
[WIP] Add 2.13.0-M5 crossbuild setting

zakpatterson:
Blocked by:

org.scalatest#scalatest_2.13.0-M5;3.0.5: not found
com.olegpy#better-monadic-for_2.13.0-M5;0.2.4: not found

PR #208, opened 2019-02-06
Remove unused monadic for plugin

zakpatterson:
This plugin is awesome, but there are no fancy fors in this project


zakpatterson:
Blocks #207

zakpatterson:
Yeah I think not merging this is fine now, but I was looking for any examples of this plugin's usage and didn't find any in this project, plus it compiles without so I don't think it's used anywhere?

PR #211, opened 2019-02-18
Extract unverified JWT

hamnis:
Motivation:
Being able to fetch the kid from the header and use that to fetch the
JWK which represents the public key and use that subsequently to verify
the JWT.


hamnis:
any chance to look at this yet?

ChristopherDavenport:
Since this is adding and not modifying we can work to get it into `.1`

hamnis:
@ChristopherDavenport any interest in pulling this in?

PR #234, opened 2019-08-03
upgrade to sbt 1.3

SethTisue:
y'all would be justified in deciding this isn't mergeable until 1.3 final is out

but in the meantime I'd like to see if it passes on your CI, that would be useful data for me in the context of some community build work I'm doing (trying to unfreeze tsec there and simplify our config tweaks)

/cc @eed3si9n

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 language: scala
 
 before_install:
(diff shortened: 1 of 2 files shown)


SethTisue:
green on 2.12, test failures on 2.11 and 2.13. I'm unfamiliar with the code, so the nature of the errors is unclear to me.

that concludes the experiment...

PR #244, opened 2019-09-13
Upgrade: cats 2.0.0

lorandszakacs:
Only non-stable library is http4s. 

Had to drop 2.11 support because both http4s and circe dropped it.


lorandszakacs:
First time I used the feature to accept a code change from github directly. Seems to work nicely 😄

